PR #2, opened 2022-04-13
create new split view component

samrichardsontylertech:
Following the pattern outlined [here](https://www.w3.org/TR/wai-aria-practices-1.1/#windowsplitter), the split view component is a container that lays out pane components along an axis. Each pane can have a handle on the leading or trailing side to enable resizing.


src/lib/split-view/split-view-panel/_mixins.scss
```diff
@@ -0,0 +1,267 @@
...
+      }
+    }
+  }
+}
+
+@mixin vertical-styles() {
+  .forge-split-view-panel[orientation=vertical] {
+    @include base-vertical;
+
+    .forge-split-view-l {
```
jake-crawford-tyler, line 76: This might be a typo, like `forge-split-view-panel`

src/lib/split-view/split-view-panel/split-view-panel-adapter.ts
```diff
@@ -0,0 +1,211 @@
...
+    this._icon = getShadowElement(component, SPLIT_VIEW_PANEL_CONSTANTS.selectors.ICON) as IIconComponent;
+    this._content = getShadowElement(component, SPLIT_VIEW_PANEL_CONSTANTS.selectors.CONTENT);
+  }
+
+  public initialize(): void {
+    // Set the parent split view
+    const parent = this._component.parentElement;
+    if (parent?.tagName.toLowerCase() === SPLIT_VIEW_CONSTANTS.elementName) {
+      this._parent = parent as ISplitViewComponent;
+    }
```
DRiFTy17, line 59: I know this will probably break if the `<forge-split-view>` is not a direct parent, but just wanted to mention we have a utility in the forge core lib called `walkUpUntil()` that we could use to locate the parent. You likely already know about it, just wanted to mention it:
https://github.com/tyler-technologies-oss/forge-core/blob/9251d3747c1711f8fde08ee436fd45fef5b8498d/src/utils/dom-utils.ts#L629-L643

src/lib/split-view/split-view/split-view-foundation.ts
```diff
@@ -0,0 +1,204 @@
...
+    const panels = this._adapter.getSlottedPanels();
+    const increment = target.position === 'end' ? 1 : -1;
+    let layer = target.position === 'end' ? SplitViewAnimatingLayer.Under : SplitViewAnimatingLayer.Above;
+
+    panels.forEach(panel => {
+      // Increment the layer if moving into or out of the target panel
+      if (panel === target || layer === SplitViewAnimatingLayer.Active) {
+        layer += increment;
+      }
+      panel.style.setProperty('--forge-split-view-animating-layer', layer.toString());
```
DRiFTy17, line 191: May be worth moving this CSS custom prop name to constants.

src/lib/split-view/split-view/split-view-foundation.ts
```diff
@@ -0,0 +1,204 @@
...
+    });
+  }
+
+  /**
+   * Removes layering after an animation.
+   */
+  public unstackSlottedPanels(): void {
+    const panels = this._adapter.getSlottedPanels();
+    panels.forEach(panel => {
+      panel.style.removeProperty('--forge-split-view-animating-layer');
```
DRiFTy17, line 201: Use constant?

src/lib/split-view/split-view-panel/split-view-panel-foundation.ts
```diff
@@ -0,0 +1,691 @@
...
+  }
+
+  /**
+   * Performs cleanup logic after a keyboard driven resize.
+   * 
+   * @param evt 
+   */
+  private _tryHandleArrowKeyUp(evt: KeyboardEvent): void {
+    if (evt.key === 'ArrowLeft' || evt.key === 'ArrowRight' || evt.key === 'ArrowUp' || evt.key === 'ArrowDown') {
+      // TODO: handle any work that should happen after resizing
```
DRiFTy17, line 281: Anything we want to take care of here right now?

src/lib/split-view/split-view/_mixins.scss
```diff
@@ -0,0 +1,42 @@
+@use './variables';
```
DRiFTy17, line 1: The `_variables.scss` file is empty, do you have anything to move in there? If not, let's just remove the file for now.

src/lib/split-view/split-view/split-view.ts
```diff
@@ -0,0 +1,90 @@
...
+  interface HTMLElementTagNameMap {
+    'forge-split-view': ISplitViewComponent;
+  }
+}
+
+@CustomElement({
+  name: SPLIT_VIEW_CONSTANTS.elementName,
+  dependencies: [SplitViewPanelComponent]
+})
+export class SplitViewComponent extends HTMLElement implements ISplitViewComponent {
```
DRiFTy17, line 28: Let's update this to extend `BaseComponent` from core to be consistent with other components (I know it doesn't technically share anything right now, but it's there for that purpose in the future).

src/lib/split-view/split-view-panel/split-view-panel.ts
```diff
@@ -0,0 +1,157 @@
...
+}
+
+@CustomElement({
+  name: SPLIT_VIEW_PANEL_CONSTANTS.elementName,
+  dependencies: [
+    IconComponent,
+    RippleComponent
+  ]
+})
+export class SplitViewPanelComponent extends HTMLElement implements ISplitViewPanelComponent {
```
DRiFTy17, line 50: Let's update this to extend `BaseComponent`

src/lib/split-view/split-view-panel/split-view-panel.ts
```diff
@@ -0,0 +1,157 @@
...
+import { SplitViewPanelAdapter } from './split-view-panel-adapter';
+import { SplitViewOrientation } from '../split-view/split-view-constants';
+import { ISplitViewBase } from '../core/split-view-base';
+import { IconComponent, IconRegistry } from '../../icon';
+import { RippleComponent } from '../../ripple';
+
+import template from './split-view-panel.html';
+import styles from './split-view-panel.scss';
+
+export interface ISplitViewPanelComponent extends ISplitViewBase, ICustomElement {
```
DRiFTy17, line 16: Instead of `ICustomElement` let's use `IBaseComponent`

src/lib/split-view/split-view/split-view.ts
```diff
@@ -0,0 +1,90 @@
...
+import { SplitViewAdapter } from './split-view-adapter';
+import { SplitViewFoundation } from './split-view-foundation';
+import { SplitViewOrientation, SPLIT_VIEW_CONSTANTS } from './split-view-constants';
+import { ISplitViewPanelComponent, SplitViewPanelComponent } from '../split-view-panel';
+import { ISplitViewBase } from '../core/split-view-base';
+
+import template from './split-view.html';
+import styles from './split-view.scss';
+
+export interface ISplitViewComponent extends ISplitViewBase, ICustomElement {
```
DRiFTy17, line 12: Instead of `ICustomElement` let's use `IBaseComponent`

src/lib/split-view/core/index.ts
```diff
@@ -0,0 +1 @@
+export * from './split-view-core-utils';
```
DRiFTy17, line 1: We may want to also export `split-view-base` from here as well?

DRiFTy17 (commented):
Really nice job on this. Code is very clean and easy to follow. UI interactions are great. I think this is a nice first iteration and should take care of our initial needs. We can always add features on top of it as needed.

I ran into a few issues with the Storybook demo, but I have a feeling it's just due to updating some properties and not others. These may be perceived as bugs or broken demos. I might suggest we separate the demos out into multiple stories with less ability to adjust all properties so that users can see common usages more easily, but of course leave the default demo with the ability to change as much as we can without it being a pain. Another option is to update the React demo to automatically set other properties when changing things like orientation to make it less confusing.

I would also like us to run some validation tests within an Angular application to help uncover anything else we may have missed.

Left a bunch of minor comments below. Once we get tests written and an Angular example completed we'll be ready to merge.

src/lib/split-view/split-view/split-view-foundation.ts
```diff
@@ -0,0 +1,204 @@
...
+  private _onPanelClose(evt: Event): void {
+    this._updateSlottedPanelsAccessibility(evt.target as ISplitViewPanelComponent);
+  }
+
+  private _onPanelOpen(evt: Event): void {
+    this._updateSlottedPanelsAccessibility(evt.target as ISplitViewPanelComponent);
+  }
+
+  private _onResize(entry: ResizeObserverEntry): void {
+    // TODO: Prompt panels to adjust size to fit
```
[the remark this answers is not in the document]
samrichardsontylertech: I need to figure out what happens when the component becomes smaller than the width of the panels. Ideally they'd resize so there's no overflow and the apparent widths are the actual widths, but it's complicated when there are multiple panels and min and max are set. I thought I might sit on it until a future iteration.

src/lib/split-view/split-view-panel/split-view-panel.html
```diff
@@ -0,0 +1,11 @@
+<template>
+  <div class="forge-split-view-panel" id="root" part="root">
+    <div class="forge-split-view-panel__handle" id="handle" part="handle" role="separator" aria-controls="content" tabindex="0">
+      <forge-icon class="forge-split-view-panel__icon" id="icon" part="icon"></forge-icon>
+      <forge-ripple part="ripple"></forge-ripple>
+    </div>
+    <div class="forge-split-view-panel__content" id="content" part="content" role="group">
```
samrichardsontylertech, line 7: I added a `group` role to the content element. This *might* be misused here but seems to do what I want — exposes it to the accessibility tree as a logical grouping of a larger widget without adding it to the page summary.

src/lib/split-view/split-view-panel/split-view-panel.ts
```diff
@@ -0,0 +1,214 @@
...
+
+import template from './split-view-panel.html';
+import styles from './split-view-panel.scss';
+
+export interface ISplitViewPanelComponent extends ISplitViewBase, IBaseComponent {
+  position: SplitViewPanelPosition;
+  size: number | string;
+  min: number;
+  max: number | undefined;
+  label: string;
```
DRiFTy17, line 22: Let's rename this to something like `accessibleLabel`/`accessible-label` to make it more clear.

src/lib/split-view/split-view-panel/split-view-panel.ts
```diff
@@ -0,0 +1,214 @@
...
+        this.disabled = coerceBoolean(newValue);
+        break;
+      case SPLIT_VIEW_PANEL_CONSTANTS.attributes.DISABLE_CLOSE:
+        this.disableClose = coerceBoolean(newValue);
+        break;
+      case SPLIT_VIEW_PANEL_CONSTANTS.attributes.AUTO_CLOSE:
+        this.autoClose = coerceBoolean(newValue);
+        break;
+    }
+  }
```
DRiFTy17, line 118: Setting default attributes on individual panels does not set the underlying property (because the panels always pull their values from the parent `<forge-split-view>` element as the default).

I tested the following attributes specifically, but others may have issues:
- `<forge-split-view-panel auto-close>`
- `<forge-split-view-panel disable>`
- `<forge-split-view-panel disable-close>`
- `<forge-split-view-panel orientation="vertical">` (this is likely an internally used property on an individual panel, correct? We may be able to just not document this at the panel level)

src/lib/split-view/split-view-panel/split-view-panel-foundation.ts
```diff
@@ -0,0 +1,689 @@
...
+  private _size: number | string = '200';
+  private _min = 0;
+  private _max: number | undefined;
+  private _label = 'Split view panel';
+  private _open = true;
+  private _disabled = false;
+  private _disableClose = false;
+  private _autoClose = false;
+
+  private _autoCloseThreshold = 0;
```
DRiFTy17, line 36: Let's add a public property for this so it can be controlled separately by a developer.

src/lib/split-view/split-view-panel/split-view-panel.ts
```diff
@@ -0,0 +1,214 @@
...
+import { SplitViewOrientation } from '../split-view/split-view-constants';
+import { ISplitViewBase } from '../core/split-view-base';
+import { IconComponent, IconRegistry } from '../../icon';
+import { RippleComponent } from '../../ripple';
+
+import template from './split-view-panel.html';
+import styles from './split-view-panel.scss';
+
+export interface ISplitViewPanelComponent extends ISplitViewBase, IBaseComponent {
+  position: SplitViewPanelPosition;
```
DRiFTy17, line 18: I'm still not sold on the name "position" for this property. If we can put some more info in the docs around it to make it clear then I think that will help. I was a bit confused when I first started using it, but then it started to make sense once I figured it out.

@nickonometry @MikeMatusz do you guys have any thoughts on this one? Sam and I chatted about it, and from his perspective it relates to 1) which side of the panel the handle goes on (which panel the handle is controlling) and 2) which direction the panel will open/close toward when the `open` property is used. We are struggling to come up with a short but descriptive name since those two things negate themselves.

My thought was that we could pick the primary thing it should represent, but I want it to be easy to grok for a new developer using the component is all.

src/stories/src/components/split-view/split-view.mdx
```diff
@@ -0,0 +1,265 @@
...
+
+</PageSection>
+
+---
+
+<PageSection>
+
+## Example
+
+### Default
```
DRiFTy17, line 27: Let's add some descriptive text for this example explaining why and how it works in this simple usage.

src/stories/src/components/split-view/split-view.mdx
```diff
@@ -0,0 +1,265 @@
...
+import { SplitViewLeadingCodeHtml } from './code/split-view-leading.ts';
+import { SplitViewVerticalCodeHtml } from './code/split-view-vertical.ts';
+import { SplitViewMultipleSplitCodeHtml } from './code/split-view-multiple-split.ts';
+import { SplitViewMultipleStackedCodeHtml } from './code/split-view-multiple-stacked.ts';
+
+<PageSection>
+
+# Split view
+
+The split view component presents `<forge-split-view-panel>` elements side-by-side as resizable panels.
```
DRiFTy17, line 17: Let's build on this overview and some more context around what it is used for, why it might be used, and when it might be used. We can add some ground rules for the component at a high level here too.

src/stories/src/components/split-view/split-view.mdx
```diff
@@ -0,0 +1,265 @@
...
+
+## Example
+
+### Default
+
+<LiveDemo layout="column" fullWidth codeHtml={SplitViewDefaultCodeHtml()}>
+  <SplitViewDemo />
+</LiveDemo>
+
+### Leading panel
```
DRiFTy17, line 33: Same as adding more context for the Default demo, let's describe what `position` is doing here so it's more clear.

Same for the rest of the examples while we're at it. It will just help clear things up before questions are asked.

src/stories/src/components/split-view/split-view.mdx
```diff
@@ -0,0 +1,265 @@
...
+| `end`                               | Resizes the panel to its maximum size.
+
+> An arrow key combined with `shift` resizes faster, ten pixels at a time. On its own an arrow key
+> resizes one pixel at a time.
+
+> The resize handle is focusable and receives all keyboard events.
+
+</PageSection>
+
+<PageSection>
```
DRiFTy17, line 204: Let's add the `CSS parts` information here. See docs pages for other components as an example.

src/lib/split-view/split-view-panel/split-view-panel.html
```diff
@@ -0,0 +1,11 @@
+<template>
+  <div class="forge-split-view-panel" id="root" part="root">
+    <div class="forge-split-view-panel__handle" id="handle" part="handle" role="separator" aria-controls="content" tabindex="0">
```
DRiFTy17, line 3: We should ensure that this element gets focus when clicking with the mouse. Right now you can only tab to it, but the default action should to set focus (can we make that cancellable too?)

Another item about the handle might be to show some feedback when the user reaches the min or max size. I'm thinking we could trigger the ripple effect, or maybe flash the bar using the danger theme color (maybe both together?). Just something to tell them they've reached the limit and it's not a bug basically.

src/lib/split-view/split-view-panel/split-view-panel.ts
```diff
@@ -0,0 +1,214 @@
...
+  size: number | string;
+  min: number;
+  max: number | undefined;
```
DRiFTy17, line 21: Since `size` accepts a percentage value, I think both `min` and `max` properties should also accept the same. Maybe we could also allow for the `px` to be specified and just parse it out? I feel as though people may provide values like `300px` or just `300` as well things like `50%`.

Which also brings me to another though, can we enable the usage of CSS custom properties for these things. It feels like a stylistic value like this should be set via CSS (although I know we calculate everything via JavaScript). Sam and I spoke about this, and it's doable, but maybe a future feature. Just wanted to note it here for others to comment.

DRiFTy17 (commented):
I set up a more realistic demo in an Angular app to validate usage there, but to also get a feel for how it is to use this component in a general application context. It works really well. I had some initial confusion around the `position` property that could use some more docs or a rename (see below), but other than that I noted a few bugs and some suggestions.

Once again, great job. Coming along nicely. This will be a nice component to add to the library.

Side note, we should also get a PR going on the [forge-react](https://github.com/tyler-technologies-oss/forge-react) repo to add these two new element wrappers. Additionally, we will need a PR for some Forge design guidance on this as well (@nickonometry for visibility).

src/lib/split-view/split-view-panel/split-view-panel.ts
samrichardsontylertech: The panel's orientation is strictly internal. I didn't expose the property itself through the API, just a `setOrientation()` method out of necessity (the parent needs to access it). The others are the only properties shared between the components so it makes sense that they're affected.

src/lib/split-view/split-view-panel/split-view-panel.ts
MikeMatusz: I really like the `resizable` concept that came out of the Slack convo for this

src/lib/split-view/split-view-panel/split-view-panel.ts
samrichardsontylertech: This should be fixed now. I changed each of the shared properties to be optional in the panel. If one is set it'll use it, otherwise the parent property will be used.

src/lib/split-view/split-view-panel/split-view-panel.ts
samrichardsontylertech: I've been thinking about this a lot. Setting size constraints through custom properties is a big ask — we'll need to be able to keep track of the min and max sizes to make the accessibility features work and I'm just not seeing a way to do that. Setting the initial size via CSS works fine, but it begins to break down when I think about what happens when the user resizes.

I'd really like to offload as much stylistic and mathematical work to CSS as possible but it'll take time to figure out how to do that without compromising anything.

DRiFTy17 (dismissed):
LGTM. The docs updates clarify the handle position much more clearly now, and I think this is a great first iteration that covers most if not all usages we will need for now. Great job on this!

DRiFTy17 (approved):
LGTM. Once you're ready to merge feel free. Let's make a PR against the React adapter lib next to add the definition there, as well as the vscode extension for snippets.

DRiFTy17 (approved):

PR #135, opened 2022-08-22
fix(select): fix a11y when navigating options via keyboard and screen reader

DRiFTy17:
## PR Checklist

- Docs have been added / updated: N/A

## Describe the new behavior?
Fixes the accessibility of the component by ensuring that options and state are announced properly by screen readers. The focus has been moved from the internal selected text element to the host element to ensure that the options are announced properly.

## Additional information
Migrated these changes from TCW:
https://github.com/tyler-technologies/tyler-components-web/pull/768

### DIFF
--- a/src/lib/select/select/select-adapter.ts
+++ b/src/lib/select/select/select-adapter.ts
@@ -66,10 +66,13 @@ export class SelectAdapter extends BaseSelectAdapter implements ISelectAdapter {
 
   public initializeAccessibility(): void {
     this._component.setAttribute('role', 'combobox');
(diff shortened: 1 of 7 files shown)


samrichardsontylertech (approved):
Works in Chrome, which is good enough for now. Let's get a fix for WebKit out soon though.

DRiFTy17:
This bug reported to Apple sounds very similar:
https://developer.apple.com/forums/thread/706788

Just noting it here to check back later. I'm not sure if it's a bug in VoiceOver or Safari yet but I'll do some more research. We're following the combobox spec properly so this will have to be adequate for now.

PR #138, opened 2022-08-30
feat(misc): custom elements manifest

MikeMatusz:
## PR Checklist

- Tests for the changes have been added: n/a
- Docs have been added / updated: n/a

## Describe the new behavior?
Adding analyzer and configuration to generate a [Custom Elements Manifest](https://github.com/webcomponents/custom-elements-manifest) and include in the package output.  Adding minimal `@tag` JSDoc required for future planned work to improve Angular integration.  Adding example of comments for slots, CSS properties, and CSS parts to one component (app-bar-search).

Marking PR as Draft because I'm investigating moving the manifest generation to the forge-cli repo.

### DIFF
--- a/src/lib/bottom-sheet/bottom-sheet.ts
+++ b/src/lib/bottom-sheet/bottom-sheet.ts
@@ -39,6 +39,11 @@ declare global {
   }
 }
 
+/**
+ * The web component class behind the `<forge-bottom-sheet>` custom element.
+ * 
+ * @tag forge-badge
+ */
 @CustomElement({
   name: BOTTOM_SHEET_CONSTANTS.elementName,
   dependencies: [BackdropComponent]
--- a/src/lib/button-toggle/button-toggle-group/button-toggle-group.ts
+++ b/src/lib/button-toggle/button-toggle-group/button-toggle-group.ts
@@ -25,6 +25,11 @@ declare global {
   }
 }
 
+/**
+ * The web component class behind the `<forge-button-toggle>` custom element.
+ * 
+ * @tag forge-button-toggle
+ */
 @CustomElement({
   name: BUTTON_TOGGLE_GROUP_CONSTANTS.elementName,
   dependencies: [ButtonToggleComponent]
--- a/src/lib/calendar/calendar-menu/calendar-menu.ts
+++ b/src/lib/calendar/calendar-menu/calendar-menu.ts
@@ -32,6 +32,11 @@ declare global {
   }
 }
 
+/**
+ * The web component class behind the `<forge-calendar-menu>` custom element.
+ * 
+ * @tag forge-calendar-menu
+ */
 @CustomElement({
   name: CALENDAR_MENU_CONSTANTS.elementName
 })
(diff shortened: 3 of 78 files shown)


src/lib/bottom-sheet/bottom-sheet.ts
```diff
@@ -39,6 +39,11 @@ declare global {
   }
 }
 
+/**
+ * The web component class behind the `<forge-bottom-sheet>` custom element.
+ * 
+ * @tag forge-badge
```
DRiFTy17, line 45: forge-bottom-sheet

src/lib/calendar/calendar-menu/calendar-menu.ts
```diff
@@ -32,6 +32,11 @@ declare global {
   }
 }
 
+/**
+ * The web component class behind the `<forge-calendar-menu>` custom element.
+ * 
+ * @tag forge-calendar-menu
+ */
```
DRiFTy17, line 39: Is there any way (or reason why) we may want to mark a component as "internal"? This component is used within the calendar to help with separation of concerns, but isn't likely to be documented or used externally unless as part of another composite component. Just curious what your thoughts are on these ones?

DRiFTy17 (commented):
Just a few minor things to fix below.

I was looking through the generated `custom-elements.json` file and I noticed it also includes all exported modules (foundation, adapter, constants, utils...etc.) which is likely more than we care about at the moment until we integrate this into the docs generation within Storybook (in which case we still may only care about the top-level component class modules). Just wanted to note that in case you wanted to adjust, but I'm fine with the extra details regardless.

src/lib/calendar/calendar-menu/calendar-menu.ts
MikeMatusz: Looks like `@internal` and `@ignore` tags are [supported](https://custom-elements-manifest.open-wc.org/analyzer/getting-started/#supported-jsdoc) so I could definitely see if that excludes it from the manifest entirely or at least includes an `"internal": true` annotation.  The analyzer also does take an `--exclude` glob as an option if necessary.

I know Compodoc (which is used under the hood by the Angular Storybook integration) uses `@ignore` to exclude something from documentation, and TypeScript uses `@internal` to exclude things from the `.d.ts` declarations file if the `--stripInternal` option is provided.  If those two are identical to the analyzer, seems like `@internal` makes the most sense.

MikeMatusz: I confirmed that `@internal` excludes the component from the manifest.

src/lib/bottom-sheet/bottom-sheet.ts
MikeMatusz: fixed

src/lib/button-toggle/button-toggle-group/button-toggle-group.ts
```diff
@@ -25,6 +25,11 @@ declare global {
   }
 }
 
+/**
+ * The web component class behind the `<forge-button-toggle>` custom element.
+ * 
+ * @tag forge-button-toggle
```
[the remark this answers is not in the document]
MikeMatusz: fixed

DRiFTy17 (approved):

PR #149, opened 2022-10-03
feat: showHourOptions for time-picker

jake-crawford-tyler:
## PR Checklist

- Docs have been added / updated: [Y]

## Describe the new behavior?
Remove hours from the time picker dropdown, so "Now" is the only menu item. This will prevent the issue where a user types in a value with minute (e.g. 11:21) then opens the dropdown, the hour rounded down (11:00) gets selected, then tabbing off removes the minutes.

## Additional information
[NWARCH-1135](https://tylerjira.tylertech.com/browse/NWARCH-1135)

### DIFF
--- a/src/lib/time-picker/time-picker-foundation.ts
+++ b/src/lib/time-picker/time-picker-foundation.ts
@@ -655,7 +655,7 @@ export class TimePickerFoundation implements ITimePickerFoundation {
 
     // Append all leading options
     if (leadingOptions.length) {
-      if (times.length > 0) {
+      if (times.length !== 0) {
         times.splice(0, 0, { label: '', value: null, divider: true });
       }
       leadingOptions.forEach((o, index) => times.splice(index, 0, o));
--- a/src/lib/time-picker/time-picker-foundation.ts
+++ b/src/lib/time-picker/time-picker-foundation.ts
@@ -545,7 +545,7 @@ export class TimePickerFoundation implements ITimePickerFoundation {
     let activeStartIndex: number | undefined;
     
     // Find closest match in list of time options and activate/select it
-    if(options.length) {
+    if (options.length) {
       if (this._value !== null) {
         const optionIndex = this._findClosestOptionIndex(this._value, selectableOptions);
         if (optionIndex >= 0) {
--- a/src/lib/time-picker/time-picker-foundation.ts
+++ b/src/lib/time-picker/time-picker-foundation.ts
@@ -642,9 +642,20 @@ export class TimePickerFoundation implements ITimePickerFoundation {
     }
 
     // Check if we need to show the Hours options
-    if (this._showHourOptions) {
-      const value: ITimePickerOptionValue = { time: null, metadata: 'now' };
-      leadingOptions.push({ label: 'Hours Options', value });
+    if (!this._showHourOptions) {
+      // while loop
+      // while(times.length > 0) {
+      //   times.pop();
+      // }
+
+      // splice time
+      times.splice(0, times.length);
+
+      // override length
+      // times.length = 0;
+
+      // Requires times to not be a constant
+      // times = [];
     }
 
     // Check for any custom provided options to prepend
--- a/src/lib/time-picker/time-picker-foundation.ts
+++ b/src/lib/time-picker/time-picker-foundation.ts
@@ -618,22 +618,25 @@ export class TimePickerFoundation implements ITimePickerFoundation {
     const times: IListDropdownOption[] = [];
     let leadingOptions: IListDropdownOption[] = [];
     
-    for (let totalMinutes = minMinutes; totalMinutes <= maxMinutes; totalMinutes += minuteStep) {
-      if (totalMinutes === TIME_PICKER_CONSTANTS.numbers.MAX_DAY_MINUTES) {
-        break;
+    if (this._showHourOptions) {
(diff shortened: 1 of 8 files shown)


src/lib/time-picker/time-picker-foundation.ts
```diff
@@ -654,6 +656,12 @@ export class TimePickerFoundation implements ITimePickerFoundation {
       leadingOptions.forEach((o, index) => times.splice(index, 0, o));
     }
 
+    // Check if we need to show the Hours options
+    if (!this._showHourOptions) {
+      // empty times array
+      times.splice(0, times.length);
+    }
```
MikeMatusz, line 663: I still think rather than clearing it out, you could just wrap [time-picker-foundation.ts#L621-L636](https://github.com/tyler-technologies-oss/forge/blob/41b2fc1b76aee1d21a32173e61ebcaab13652298/src/lib/time-picker/time-picker-foundation.ts#L621-L636) in `if (this._showHourOptions)` and then let Now and the custom options just get added to the empty array.  Just seems better to not do the work, then do it and then undo it.

Might also need to exclude the divider on line 655 if `times.length === 0`.

src/lib/time-picker/time-picker.ts
```diff
@@ -217,6 +222,9 @@ export class TimePickerComponent extends BaseComponent implements ITimePickerCom
   @FoundationProperty()
   public showNow: boolean;
 
+  @FoundationProperty()
+  public showHourOptions: boolean;
```
MikeMatusz, line 226: I know the rest of this file doesn't have one, but might be nice to add some JSDoc to this property with a description.  For now you'd probably want to duplicate it in `time-picker-args.ts`, but the comment here would get picked up by the Custom Elements Manifest and could eventually be used to generate the documentation.

src/test/spec/time-picker/time-picker.spec.ts
```diff
@@ -459,6 +459,12 @@ describe('TimePickerComponent', function(this: ITestContext) {
     expect(this.context.component.showNow).toBeTrue();
   });
 
+  it('should set showHourOptions via attribute', function(this: ITestContext) {
```
MikeMatusz, line 462: This test is good, but you should probably also add one to check that they're actually excluded from the options, similar to the ['should show "now" as the first option'](https://github.com/tyler-technologies-oss/forge/blob/138f9291d78da4fc2717a277b1cdfff2120cdba5/src/test/spec/time-picker/time-picker.spec.ts#L952-L964) test but verifying that "now" is the only option if this property is false.

src/lib/time-picker/time-picker-foundation.ts
jake-crawford-tyler: Both fixed with most recent push

src/test/spec/time-picker/time-picker.spec.ts
jake-crawford-tyler: thanks, added 2 additional tests.

DRiFTy17 (commented):
The changes are looking good to me. I found one small bug that occurs now that there is the potential for no options to be available in the dropdown. This is an edge case for sure, but still something I think we should take care of.

1. Set `showHourOptions` to `true`
2. Open time-picker dropdown and select a time from the existing hour options
3. Set `showHourOptions` to `false`
4. Open the dropdown and see following error in dev console

image

Looks like we need to add a safety check somewhere if there are no options found to select.

madhater (approved):

src/lib/time-picker/time-picker-foundation.ts
```diff
@@ -529,7 +531,7 @@ export class TimePickerFoundation implements ITimePickerFoundation {
   }
 
   private _openDropdown(): void {
-    if (!this.allowDropdown) {
+    if (!this.allowDropdown || this._generateTimeOptions.length) {
```
madhater, line 534: _generateTimeOptions is a function, it will always return false here, so isn't accomplishing anything. What could work is instead is changing this to:
const options = this._generateTimeOptions();  // move from line 542
if(!this.allowDropDown || options.length === 0) {
  return;
}

madhater: actually, i'll revise the above comment, i think it should be:

if (!this.allowDropdown || (!this.showHourOptions && !this.showNow)) {
  return;
}

showHourOptions == false && this.showNow == false are equivalent to not allow drop down as there will be nothing in the drop down to show. If either are true, there should be options in the drop down to show, and worth showing it.

This will remove the later changes in this function below I think.

src/lib/time-picker/time-picker-foundation.ts
```diff
@@ -588,10 +592,14 @@ export class TimePickerFoundation implements ITimePickerFoundation {
   }
 
   private _findClosestOptionIndex(value: number, options: Array<IListDropdownOption<ITimePickerOptionValue>>): number {
-    const closestItem = options.reduce((prev, curr) => {
+    if (options.length) {
```
madhater, line 595: This seems uneccessary as the two places it's called from should prevent this method from being executed if options.length is 0

jake-crawford-tyler: fixed with most recent push

src/lib/time-picker/time-picker-foundation.ts
jake-crawford-tyler: After discussion, we decided on testing options' length. This has been pushed for retesting

DRiFTy17 (approved):

PR #160, opened 2022-10-14
feat: expand nullable types to allow undefined

MikeMatusz:
## PR Checklist

- Tests for the changes have been added: n/a

## Describe the new behavior?
* The types of all nullable properties of components have been expanded to allow `undefined`, along with the associated foundation properties where necessary.
* Autocomplete `filter` and `optionBuilder` properties have been made nullable since they were already null-checked and functioning this way.
* Chip component `value` has been changed from `string` to `any` to match the foundation property.
* Strict `===` or `!==` checks against `null` or `undefined` have been replaced with equivalent double-equal operations to encompass both where appropriate.

## Additional information
These issues were encountered when testing using Angular proxy components instead of CUSTOM_ELEMENTS_SCHEMA, with strict template type-checking enabled.  These type expansions/corrections will provide a better experience for Angular developers once those proxy components become available, and also any TypeScript developers setting the properties on the components directly.


DRiFTy17:
Are there any linting rules we can use to ensure that the code stays in compliance as we move forward? If not, maybe we can create one specific to our needs.

src/lib/chips/chip/chip-foundation.ts
```diff
@@ -8,7 +8,7 @@ export interface IChipFoundation extends ICustomElementFoundation {
   selected: boolean;
   disabled: boolean;
   invalid: boolean;
-  value: string;
+  value: any;
```
samrichardsontylertech, line 11: Not that this should be addressed here and now but have we given any thought to phasing out `any` and replacing it with `unknown` throughout the library?

MikeMatusz: I'd be all in favor of that, and almost used it for this one but since it was already `any` in some places didn't want to rock the boat with this PR.  With the introduction of `unknown` there are very few places `any` is actually needed anymore.

MikeMatusz:
> Are there any linting rules we can use to ensure that the code stays in compliance as we move forward? If not, maybe we can create one specific to our needs.

Some good discussion on typescript-eslint/typescript-eslint#821 related to this.  Personally, I'd love to centralize on `undefined` rather than `null`, though Angular does complicate things by defaulting to `null` in many cases (and isn't willing to change: angular/angular#42713) and some native APIs can return null.  Possibly too opinionated for Forge, though.  I don't see any existing rules that enforce always allowing both `null | undefined` with explicit types, though.  We could also create a utility type `type Nullable<T> = T | null | undefined` (sometimes called `Maybe`) and try to enforce that instead of null/undefined.

I think we'd also want to make sure our `eqeqeq` lint rule is configured to either `smart` or `always` with `null: 'never'` so our null checks always account for both.

DRiFTy17:
> We could also create a utility type type Nullable<T> = T | null | undefined (sometimes called Maybe) and try to enforce that instead of null/undefined.

I'm all for that. We should always allow both since they are typically treated the same in the case of our properties.

> I think we'd also want to make sure our eqeqeq lint rule is configured to either smart or always with null: 'never' so our null checks always account for both.

Agreed on this as well for branching logic.

samrichardsontylertech (approved):

PR #181, opened 2022-11-03
feat: added property on base field component to allows the addon end slot to remain enabled when the field is otherwise disabled

Alex-Oxthorn:
## PR Checklist

- Docs have been added / updated: [Y]

## Describe the new behavior?
adds new attribute/property to base field component that will allow the addon-end slot content to remain enabled when the field is otherwise disabled.

### DIFF
--- a/src/lib/field/_selector.scss
+++ b/src/lib/field/_selector.scss
@@ -1,5 +1,6 @@
 @use 'sass:list';
 @use './base';
+@use '../theme';
 
 @mixin field($component-name: 'forge-field', $exclude: ()) {
   // Core
@@ -527,7 +528,7 @@
         }
       }
     }
-    &.forge-field--disabled {
+    &.forge-field--disabled:not(.forge-field__addon-end-container--enabled) {
       .forge-field__addon-end-container {
         @include base.addon-end-border-left-color(disabled);
       }
@@ -548,6 +549,25 @@
     &--disabled:not(.forge-field__addon-end-container--enabled) {
       ::slotted([slot='addon-end']) {
         @include base.icon-color(disabled);
+      }
+    }
+  }
+
+  // Color
+  @if list.index($exclude, forge-field__addon-end-container--enabled) == null {
+    .forge-field__addon-end-container--enabled {
+      .forge-field__addon-end-container {
+
+        @include theme.property(background-color, var(--mdc-theme-surface));
+        @include theme.property(color, var(--mdc-theme-text-primary-on-background));
+        @include theme.property(border-color, border-color);
+        border-width: 1px;
+        border-style: solid;
+        border-top-right-radius: 4px;
+        border-bottom-right-radius: 4px;
+        box-sizing: border-box;
+        -moz-box-sizing: border-box;
+        -webkit-box-sizing: border-box;
       }
     }
   }
--- a/src/lib/field/_selector.scss
+++ b/src/lib/field/_selector.scss
@@ -564,8 +564,6 @@
       border-top-right-radius: 4px;
       border-bottom-right-radius: 4px;
       box-sizing: border-box;
-      -moz-box-sizing: border-box;
-      -webkit-box-sizing: border-box;
     }
   }
 }
--- a/src/lib/field/_selector.scss
+++ b/src/lib/field/_selector.scss
@@ -553,22 +553,19 @@
     }
   }
 
-  // Color
-  @if list.index($exclude, forge-field__addon-end-container--enabled) == null {
-    .forge-field__addon-end-container--enabled {
-      .forge-field__addon-end-container {
-
-        @include theme.property(background-color, var(--mdc-theme-surface));
-        @include theme.property(color, var(--mdc-theme-text-primary-on-background));
-        @include theme.property(border-color, border-color);
-        border-width: 1px;
-        border-style: solid;
-        border-top-right-radius: 4px;
-        border-bottom-right-radius: 4px;
-        box-sizing: border-box;
-        -moz-box-sizing: border-box;
-        -webkit-box-sizing: border-box;
-      }
+  // Addon End Slot Always Enabled.
+  &.forge-field__addon-end-container--enabled.forge-field--disabled:not(.forge-field--invalid) {
+    .forge-field__addon-end-container {
+      @include theme.property(background-color, surface);
+      @include theme.property(color, text-primary-on-background);
+      @include theme.property(border-color, border-color);
+      border-style: solid;
+      border-width: 1px;
+      border-top-right-radius: 4px;
+      border-bottom-right-radius: 4px;
+      box-sizing: border-box;
+      -moz-box-sizing: border-box;
+      -webkit-box-sizing: border-box;
     }
   }
 }
--- a/src/lib/field/_selector.scss
+++ b/src/lib/field/_selector.scss
@@ -545,7 +545,7 @@
         @include base.icon-color(invalid);
       }
     }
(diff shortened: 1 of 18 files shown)


src/lib/field/field.ts
```diff
@@ -1,14 +1,16 @@
...
 import { FieldFoundation } from './field-foundation';
-import { coerceBoolean, ensureChildren, FoundationProperty } from '@tylertech/forge-core';
 
 export interface IFieldComponent extends IBaseComponent {
   density: FieldDensityType;
   floatLabelType: FieldFloatLabelType;
   shape: FieldShapeType;
   invalid: boolean;
   required: boolean;
+  addonEndAlwaysEnabled: boolean;
```
DRiFTy17, line 13: I'm not in love with this name, but I also can't think of anything better. Naming is hard... haha

I assume you went through the same thought process and this seems reasonable, but I'll think on it for a min and see if anything comes to mind otherwise we can go with this.

src/lib/field/_selector.scss
```diff
@@ -545,7 +545,7 @@
         @include base.icon-color(invalid);
       }
     }
-    &--disabled {
+    &--disabled:not(.forge-field__addon-end-container--enabled) {
```
DRiFTy17, line 548: Is there desire to not have the disabled background color behind this slot?

I ask because it's not immediately obvious that the element is interactive as can be seen here:
image

The current DOM structure and styles wouldn't support that too well as I'm sure you noticed, but I'm curious if that was the end goal of the design and the complexity of that change was too high or if this is all we needed to support your usage?

src/test/spec/text-field/text-field.spec.ts
```diff
@@ -237,39 +230,39 @@ describe('TextFieldComponent', function(this: ITestContext) {
       expectFloatingLabelState(this.context, false);
     });
 
-    it('should set dense via property', function(this: ITestContext) {
+    it('should set dense via property', function (this: ITestContext) {
```
DRiFTy17, line 233: Do you have a formatter that is updating these files? It may be worth disabling that to reduce noise.

src/test/spec/text-field/text-field.spec.ts
```diff
@@ -518,117 +511,155 @@ describe('TextFieldComponent', function(this: ITestContext) {
     });
   });
 
-  describe('with leading icon', function(this: ITestContext) {
-    beforeEach(function(this: ITestContext) {
+  describe('API - addon-end-always-enabled', function (this: ITestContext) {
```
DRiFTy17, line 514: Do we want to add a test that actually attempt to click on an element within this slot when this is enabled?

Alex-Oxthorn: I don't think its strictly needed personally.

src/test/spec/text-field/text-field.spec.ts
Alex-Oxthorn: I do, and I thought I had nuked all of those before committing, guess I missed some.

src/lib/field/field.ts
Alex-Oxthorn: Yea, I ran through a couple different names and this seemed to be the least terrible. As you said, naming stuff is hard.

src/test/spec/text-field/text-field.spec.ts
DRiFTy17: My only concern is that without it we _could_ have a regression where the classes are still added fine, but the element in this slot ends up being non-interactive.

Alex-Oxthorn: The only reason they were weren't interactable before the css that prevented cursor actions.

src/lib/field/_selector.scss
```diff
@@ -545,11 +546,27 @@
...
+      @include theme.property(background-color, surface);
+      @include theme.property(color, text-primary-on-background);
+      @include theme.property(border-color, border-color);
+      border-style: solid;
+      border-width: 1px;
+      border-top-right-radius: 4px;
+      border-bottom-right-radius: 4px;
+      box-sizing: border-box;
+      -moz-box-sizing: border-box;
+      -webkit-box-sizing: border-box;
```
DRiFTy17, line 568: Just an FYI, we run CSS through autoprefixer and these will be added automatically.

src/lib/field/_selector.scss
```diff
@@ -545,11 +546,27 @@
...
       ::slotted([slot='addon-end']) {
         @include base.icon-color(disabled);
       }
     }
   }
+
+  // Addon End Slot Always Enabled.
+  &.forge-field__addon-end-container--enabled.forge-field--disabled:not(.forge-field--invalid) {
+    .forge-field__addon-end-container {
+      @include theme.property(background-color, surface);
```
DRiFTy17, line 559: While I understand that we need this background override due to our current DOM structure, it does go against the original intent of the component to inherit the container background when enabled. Not a big deal, and someone could override per their needs if they run into it.

Just wanted to call it out so the expectation is clear, and so we can discuss when we look at this for v3.

DRiFTy17 (commented):
Overall things are looking good to me, but we will want to enable this functionality in the select component as well. It does not currently derive from the `FieldComponent` class so we'll need to manually add the property there...

src/lib/field/_selector.scss
jake-crawford-tyler: Good point, this is a relic of some in browser testing, so we should removed this @Alex-Oxthorn

src/lib/field/_selector.scss
jake-crawford-tyler: @DRiFTy17 just to clarify the DOM positioning of this doesn't allow us to use transparent as the color here as the background color behind the add on would show through, which is the disabled. I agree the layering isn't ideal but this seems like the correct solution for the moment. Definitely worth a discussion for v3

jake-crawford-tyler (changes requested):
Small change needed to remove the cross browser code, then we are good.

src/lib/field/_selector.scss
Alex-Oxthorn: Are you just documenting the intent here @DRiFTy17 , or is there something that you would like changed about this?

DRiFTy17:
> Small change needed to remove the cross browser code, then we are good.

@Alex-Oxthorn @jake-crawford-tyler Are you guys planning to update the select component within this PR or separately?

Alex-Oxthorn:
Its not something we need atm, so it would be a separate effort if at all.

DRiFTy17:
> Its not something we need atm, so it would be a separate effort if at all.

Ok, I'll just go ahead and push the changes to the select and update this PR. I'd prefer to have the text-field and select stay in sync as much as possible to keep expectations straight. In the future these will share more code than they do so this won't be as much of a concern.

DRiFTy17:
@Alex-Oxthorn I've been doing some more thinking about this. I would like to run this by the Forge governance board before we merge it. I say this because I want to make sure this is a pattern we actually _want_ to support at a core level,  given that we could apply these changes just in Public Safety if necessary within the PSDF.

The original intention with these fields and slots was that the disabled state applies to the field as a whole, which encompasses the slots and their content.

Something still doesn't quite feel right about adding this oddly named property to just override some styles for a specific use case unless Forge prescribes it. I'd prefer that we instead just have you add these styles via CSS part selectors (if that's difficult currently, then we can make that easier).

I'll still update this PR with the select changes so we have it readily available.

PR #204, opened 2022-12-16
[next][dialog] static usage implementation

DRiFTy17:
- Initial PoC implementation of using `<forge-dialog>` statically
- Still supports dynamic usage as well
- No breaking changes introduced


DRiFTy17:
The Storybook demo will not work until we update the forge-react lib to append the `<forge-dialog>` element to the DOM, but it also needs to be updated in general to support the static usage expectation anyway.

PR #212, opened 2023-01-11
fix(select): fixed a bug where the UI was not reflecting the selected value if the option values are changed after the component value is set

DRiFTy17:
## PR Checklist

## Describe the new behavior?
The select component was not updating its selection state if child `<forge-option>` elements change their `value` property. This was causing issues where a value being set on the `<forge-select>` would not reflect in the UI if the `<forge-option>` values are set **after** the `<forge-select>` value...

The new behavior is that the `<forge-option>` elements will now dispatch a `forge-option-value-change` event when the `value` setter is executed. This is intended to be an internal use only event that will not be documented. This event allows for the parent `<forge-select>` element to capture that event and ensure that it updates its selection state internally and reflects it properly in the UI.

## Additional information
This issue was discovered in an Angular app where incremental rendering and the order of bindings was causing the `<forge-select>` element to update its value before the `<forge-option>` element `value` bindings were set.

### DIFF
--- a/src/lib/select/option/option-foundation.ts
+++ b/src/lib/select/option/option-foundation.ts
@@ -24,12 +24,13 @@ export class OptionFoundation implements IOptionFoundation {
   constructor(private _adapter: IOptionAdapter) {}
 
   /** Gets/sets the value of this option. */
-  public get value(): string {
+  public get value(): any {
     return this._value;
   }
-  public set value(value: string) {
+  public set value(value: any) {
     if (this._value !== value) {
       this._value = value;
+      this._adapter.emitHostEvent(OPTION_CONSTANTS.events.VALUE_CHANGE, this._value);
     }
   }
 
--- a/src/lib/select/core/base-select-adapter.ts
+++ b/src/lib/select/core/base-select-adapter.ts
@@ -146,9 +146,21 @@ export abstract class BaseSelectAdapter extends BaseAdapter<IBaseSelectComponent
   }
 
   public setOptionsListener(listener: (options: ISelectOption[] | ISelectOptionGroup[]) => void): SelectOptionListenerDestructor {
+    // Watch for option value property changes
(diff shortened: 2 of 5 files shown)


src/lib/select/option/option-foundation.ts
```diff
@@ -24,12 +24,13 @@ export class OptionFoundation implements IOptionFoundation {
...
   /** Gets/sets the value of this option. */
-  public get value(): string {
+  public get value(): any {
     return this._value;
   }
-  public set value(value: string) {
+  public set value(value: any) {
     if (this._value !== value) {
       this._value = value;
+      this._adapter.emitHostEvent(OPTION_CONSTANTS.events.VALUE_CHANGE, this._value);
```
DRiFTy17, line 33: Just calling this out because I don't necessarily love dispatching an event from a setter, but aside from monkey patching the `value` property on all `<forge-option>` elements, this seemed like the best approach. Especially given that this is an internal use only event that is used for notification purposes. Open to alternative suggestions.

src/lib/select/option/option-foundation.ts
```diff
@@ -24,12 +24,13 @@ export class OptionFoundation implements IOptionFoundation {
   constructor(private _adapter: IOptionAdapter) {}
 
   /** Gets/sets the value of this option. */
-  public get value(): string {
+  public get value(): any {
```
samrichardsontylertech, line 27: We should probably transition to `unknown` at some point

samrichardsontylertech (approved):

src/lib/select/option/option-foundation.ts
DRiFTy17: Agreed. I'll add that to the 3.0 roadmap. This was just a mismatching type because everywhere else that interfaces with this uses `any` currently...

PR #225, opened 2023-02-07
fix(list-item): fixed a bug where the UI could incorrectly reflect the component state

Alanoll:
## PR Checklist

- Docs have been added / updated: [N]

## Describe the new behavior?
The `forge-list-item` component will now properly toggle it's child checkbox component on initialization if the parent `forge-list` has an initial `selectedValue` provided. It does this by attempting to call the tryToggleCheckboxRadio when trySelect is called.

## Additional information
The `forge-list-item` component was not updating it's child checkbox state when Angular application was databinding the `selectedValue` and a `forge-list` component and it's `forge-list-item` components were dynamically rendered in a ngFor.

### DIFF
--- a/src/lib/list/list-item/list-item-foundation.ts
+++ b/src/lib/list/list-item/list-item-foundation.ts
@@ -69,7 +69,10 @@ export class ListItemFoundation implements IListItemFoundation {
       this._adapter.setTwoLine(this._twoLine);
     }
 
-    this._selected = this._adapter.trySelect(this._value);
+    const isSelected = this._adapter.trySelect(this._value);
+    if(isSelected != null) {
+      this._selected = isSelected;
+    }
   }
 
   public disconnect(): void {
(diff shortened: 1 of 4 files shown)


src/lib/list/list-item/list-item-foundation.ts
```diff
@@ -69,7 +69,10 @@ export class ListItemFoundation implements IListItemFoundation {
       this._adapter.setTwoLine(this._twoLine);
     }
 
-    this._adapter.trySelect(this._value);
+    const isSelected = this._adapter.trySelect(this._value);
+    if(isSelected != null) {
```
Alanoll, line 73: I added this because it's possible for the `_selected` field would be out of sync as well after initialization as well as in the `value` setter.

DRiFTy17 (commented):
The code changes look good to me.

I found one small issue though. If the `<forge-list-item>` is `selected` **before** adding to the `<forge-list>`, it looks like it properly unchecks the checkbox, but the visual selected state stays applied:
image

It's definitely an unlikely edge case, but if you don't mind could you take a look at that one while you're at it?

Alanoll:
Looks like it was the same area of code to check for fix your issue @DRiFTy17 . Pushed a fix for it to this branch.

DRiFTy17 (approved):
LGTM

PR #231, opened 2023-02-10
perf(list-item): delay ripple creation

### LINKED ISSUE #230: [list-item] performance bottleneck when initializing ripple
**Describe the bug:**
When rendering the list items for an autocomplete, there's a noticeable delay when opening the dropdown due to each item instantiating a ripple instance.  In most applications the delay is pretty minimal, but in a hybrid AngularJS app, it ends up triggering a lot of extra `$digest` cycles and the delay is often 2/3 of a second or more with 50-100 items (infinite scrolling is already in place to limit the initial amount shown).

See related issue #110 

**To Reproduce:**
Create a forge autocomplete with a large number of items (>100).
Click the button on the screen to render the elements.
Observe the render-blocking performance bottleneck

**Expected behavior:**
While generally it's desirable to keep the number of items in an autocomplete as low as possible, when some of these can be used with customer-defined data sources, this is not always possible, so the list-item should not instantiate the internal ripple interaction until it is needed to avoid this bottleneck. Developers should also prefer to use other alternatives such as pagination or infinite scrolling whenever rendering large amounts of DOM nodes, though in this particular instance, the app is already limiting the number of items and still experiencing a delay.

Deferring the instantiation of the ripple greatly improves the performance in initial testing, using the same method as on PR #115 

**Screenshots:**
N/A

-@tylertech/forge version: ^2.0.0

I have searched existing issues before creating this report? Y
Browser: All
Platform: All
OS: All

**Additional context:**
Also discussed https://github.com/tyler-technologies-oss/forge/discussions/77#discussioncomment-3109043 in regards to the @next version.

MikeMatusz:
Fixes #230 

## PR Checklist

- Docs have been added / updated: n/a

## Describe the new behavior?
A clear and concise description of the changes, including an breaking changes (if applicable)

## Additional information
Required a slight modification to the pattern used in #115 for icon-button because list-item has a `ripple` property that allows the ripple to be turned on and off.  I suspect it's an extreme edge case for this property to be set after initialization, but I think the logic should ensure a consistent state regardless.

### DIFF
--- a/src/lib/list/list-item/list-item-adapter.ts
+++ b/src/lib/list/list-item/list-item-adapter.ts
@@ -1,5 +1,6 @@
 import { addClass, getShadowElement, removeClass, requireParent, isDeepEqual, toggleClass } from '@tylertech/forge-core';
 import { BaseAdapter, IBaseAdapter } from '../../core/base/base-adapter';
+import { userInteractionListener } from '../../core/utils';
 import { IListComponent } from '../list/list';
 import { LIST_CONSTANTS } from '../list/list-constants';
 import { IListItemComponent } from './list-item';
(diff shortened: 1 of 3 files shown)


DRiFTy17 (approved):
LGTM

PR #236, opened 2023-02-12
fix(date-picker): fixed a bug where a 4-digit year was not correctly being masked with a leading 0

### LINKED ISSUE #235: [Date-Picker]: Date picker allows a 4-digit year with a leading 0
**Describe the bug:**
When inputting a date in DD/MM/YYYY format with YYYY with a leading 0, parseDateString will attempt to remove the leading zero and show a 3-digit date which also triggers isValidYearLength to return false. When clicking back into the field it triggers parseDateString again which then corrects the invalid date by adding a 0 to the end because it is now a 3-digit year.

**To Reproduce:**
1. Enable date masking
2. Set allowInvalidDate to false
3. Enter Example date 01/01/0111

**Expected behavior:**
The date should be masked instead of allowing a 3-digit date to show.

**Screenshots:**
image

 - Forge version: v2.13.1
 - Browser: All
 - Platform: All
 - OS: All

Lillious:
## PR Checklist

- Tests for the changes have been added: N/A
- Docs have been added / updated: N/A

## Describe the new behavior?
Created a conditional that removes the leading zero from a 4-digit year and allows it to be corrected through the 3-digit year conditional.

## Additional information
Fixes #235


src/lib/core/utils/date-utils.ts
```diff
@@ -47,6 +47,11 @@ export function parseDateString(value: string): Date | null {
     day = '1';
   }
 
+  // Trap for the case where a 4 digit year is entered with a leading 0 
+  if (typeof year === 'string' && year.length === 4) {
+    year = year.replace(/^0+/, '');
```
MikeMatusz, line 52: In testing, this switches the leading zero in the year to a trailing zero when tabbing off.  Was that the intent?  If a user enters `0111`, I'm not sure it's a safe assumption that they meant `1110`, and changing it automatically seems like it could be problematic, especially if they don't notice because they're already focused on the next field.

I don't have any use cases where a three-digit year would be valid, so it's possible the only time this would be encountered could be a typo, but I think it would be better UX to handle that with form validation and make the user correct the value themselves since only they know what they intended, rather than altering it this way.  Feel free to loop me into a discussion if you disagree.

MikeMatusz (changes requested):

src/lib/core/utils/date-utils.ts
Lillious: While debugging and reviewing the values, this is what appeared to happen in the backend, the only difference being that it was not visually reflecting the changes. Essentially the value would be 1110 instead of 0111 behind the scenes but would actually be showing 111 visually to the user. The goal was to match the visual with what the value actually is behind the scenes.

DRiFTy17: I agree with @MikeMatusz on this one. The value of `1110` and `0111` are two vastly different years, and both are valid dates.

The problem here is how you can't represent a number type with a leading zero of course. I believe the problem is located on the following line:
https://github.com/tyler-technologies-oss/forge/blob/3dd1057fae3e666418d38b4783bd27d7833fad80/src/lib/core/utils/date-utils.ts#L102

If we adjust that to the following, then I think that would also take care of it without adjusting the original value to something unexpected:
```typescript
const year = String(date.getFullYear()).padStart(4, '0');
```

Lillious: I updated the code with your recommended changes, and it appears to work better and now visualizes the correct value to the user instead of letting it be handled by the 3-digit year conditional evaluation statement.

Lillious: Regarding the `parseDateString()` padding that gets applied. An idea could be to utilize the same method used to determine the current century for 2-digit years, for 3-digit years as well. Instead of padding a zero at the beginning, this would be more in-line with how 2-digit years are calculated for consistency. Any thoughts on this?

DRiFTy17: I like this idea because it's consistent with how we handle 2-digit years. The only question I have is whether it is obvious enough when we change the value on them. It can be confusing to users if they type one thing, blur the field, and it changes to something completely different without them potentially even noticing...

With that being said, I still think 3-digit years are most likely either a typo, tabbed out of the field too quickly, or the user really did intend to type a 3-digit year.

It sounds like we have 3 options:

1. Leave it as-is where we pad a trailing zero: "111" -> "1110"
2. Adjust to pad leading zeros instead: "111" -> "0111"
3. Infer the current century similar to how we handle 2-digit years: "111" -> "2111"

I'm curious what others think about the "correct" solution? @MikeMatusz @jayexprs @nickonometry

Option 2 above seems to be the least destructive by keeping the value they typed exactly the same, but at the same time it's probably not what they were intending and therefore may be the least useful. I don't think option 3 is very obvious to me the same way a 2-digit year is, but I like it for consistency. Option 1 seems safe, but also it's just a graceful fallback "guess" since we don't know what they were intending...

MikeMatusz: Tagging @robsturgill-tyler too.  I'm not sure how likely any of these are to match the user's intent, so part of me thinks either it shouldn't coerce at all and just be an invalid date.  Like we support YY or YYYY, but not three digits.  These kind of edge cases are probably my least favorite UX problem because we want to be helpful but it can backfire.

That said, if we can't just keep it invalid, I still think option 2 is safest, since we accept their entry as intended and hope there's a minDate protecting them if it shouldn't be.

Option 1 feels most risky because you're more likely to end up with a far future date, or even one subtly "close enough" but not right, like if you typo 2023 as 202 or 203.  In Public Safety we pretty much always have a minDate, and even commonly (but not always) warn for dates more than 100 years in the future, so the autocorrects most likely to evade detection are the ones that are close to correct.

Option 3 is tempting, but I worry it would run into the same trap as Option 1 of being close enough to evade sanity checks while still not being what they intended.

MikeMatusz: I also agree that if we're going with option 2, we should update the line 52 to `padStart` to match as Kieran suggested above: https://github.com/tyler-technologies-oss/forge/pull/236#discussion_r1104555038

Lillious: When testing this, it only appeared to happen when 0111 got updated to 111. I think it'd be safe to assume the user intent is 0111. I wasn't able to input a strict 3-digit year on its own. With that in mind, I also agree with option 2 being the best route. Thoughts on this or any edge cases I may have missed?

MikeMatusz: The case I'm referring to is if I type in  `01/01/111` and tab off, the year changes to `1110`, which seems inconsistent now.  Kieran suggested that change above.  If we want to keep this PR narrowly scoped, I can create a separate issue for that and follow up.

MikeMatusz (approved):

src/lib/core/utils/date-utils.ts
DRiFTy17: Yeah I think we should go ahead and adjust line 52 that I mentioned to use `padStart()` for consistency in this PR. Otherwise you'll get leading zeros in the formatting, but trailing zeros in the parsing which produce different results.

DRiFTy17 (approved):

MikeMatusz (approved):

PR #250, opened 2023-03-23
fix(expansion-panel): avoid nested buttons in header

### LINKED ISSUE #247: [expansion panel] fix accessibility issues caused by nested interactive elements
**Describe the bug:**
The expansion panel header has a hardcoded `role` of `button`, which renders nested interactive elements inaccessible. Including buttons with extra functionality or more complex text content within an expansion panel header is a common pattern which may require consumers to handle certain accessibility features themselves but is not possible in the current component.

**To Reproduce:**
1. Place a button within an expansion panel header
2. Run an accessibility audit tool such as axe

**Expected behavior:**
It should be possible to place arbitrary content in an expansion panel header without rendering it inaccessible.

samrichardsontylertech:
## PR Checklist

## Describe the new behavior?
Fixes #247 

The button role and related ARIA attributes have been removed from the expansion panel header itself and a visually hidden button is appended to the end of the header. The click listener remains on the header so a click anywhere still toggles the panel, but accessibility and keyboard interactions are handled by the button.

For added flexibility a `type` property has also been added. When set to `'button'` the new button is added along with the click listener and accessibility attributes. When set to `'manual'` the expansion panel is not directly interactive and doesn't handle accessibility at all, to enable programmatic toggling through external UI elements.

## Additional information
This also fixes an issue in Safari where mousing over a panel header doesn't set the correct cursor until it's clicked.

### DIFF
--- a/src/lib/expansion-panel/expansion-panel-adapter.ts
+++ b/src/lib/expansion-panel/expansion-panel-adapter.ts
@@ -1,17 +1,19 @@
 import { getShadowElement } from '@tylertech/forge-core';
 import { BaseAdapter, IBaseAdapter } from '../core/base/base-adapter';
 import { OpenIconComponent } from '../open-icon';
+import { ForgeRipple } from '../ripple';
 import { IExpansionPanelComponent } from './expansion-panel';
-import { EXPANSION_PANEL_CONSTANTS } from './expansion-panel-constants';
+import { ExpansionPanelType, EXPANSION_PANEL_CONSTANTS } from './expansion-panel-constants';
+import { createButtonElement } from './expansion-panel-utils';
 
 export interface IExpansionPanelAdapter extends IBaseAdapter {
   initialize: (open: boolean, orientation: string) => void;
   setHeaderVisibility: (visible: boolean) => void;
   setOpenState: (open: boolean, orientation: string, animate: boolean) => void;
+  setType: (type: ExpansionPanelType, open: boolean) => void;
+  setButtonLabel: (label: string) => void;
   registerClickListener: (listener: (evt: MouseEvent) => void) => void;
   deregisterClickListener: (listener: (evt: MouseEvent) => void) => void;
-  registerKeydownListener: (listener: (evt: KeyboardEvent) => void) => void;
-  deregisterKeydownListener: (listener: (evt: KeyboardEvent) => void) => void;
   registerHeaderSlotListener: (listener: (evt: Event) => void) => void;
   deregisterHeaderSlotListener: (listener: (evt: Event) => void) => void;
 }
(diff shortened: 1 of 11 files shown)


samrichardsontylertech:
The ripple effect isn't completely working yet. The active state and animation aren't triggering at all.

src/lib/expansion-panel/expansion-panel.ts
```diff
@@ -109,6 +119,14 @@ export class ExpansionPanelComponent extends BaseComponent implements IExpansion
   @FoundationProperty()
   public declare useAnimations: boolean;
 
+  /** Sets whether an internal button and associated interactions are attached. */
+  @FoundationProperty()
+  public declare type: ExpansionPanelType;
+
+  /** Gets/sets the internal toggle button's label. */
+  @FoundationProperty()
+  public declare accessibleLabel: string;
```
MikeMatusz, line 128: Do these class fields actually need the `declare` keyword?  I thought that was only for [providing a more specific typing](https://www.typescriptlang.org/docs/handbook/2/classes.html#type-only-field-declarations) on fields inherited from a base class, but let me know if there's another usage I'm not aware of.

src/lib/expansion-panel/expansion-panel-adapter.ts
```diff
@@ -20,12 +22,15 @@ export class ExpansionPanelAdapter extends BaseAdapter<IExpansionPanelComponent>
...
   private _headerSlotElement: HTMLSlotElement;
+  private _buttonElement: HTMLButtonElement | undefined;
+  private _ripple: ForgeRipple | undefined;
 
   constructor(component: IExpansionPanelComponent) {
     super(component);
     this._headerElement = getShadowElement(this._component, EXPANSION_PANEL_CONSTANTS.selectors.HEADER);
     this._contentElement = getShadowElement(this._component, EXPANSION_PANEL_CONSTANTS.selectors.CONTENT);
     this._headerSlotElement = getShadowElement(this._component, EXPANSION_PANEL_CONSTANTS.selectors.HEADER_SLOT) as HTMLSlotElement;
+    this._ripple = new ForgeRipple(this._headerElement);
```
MikeMatusz, line 33: Haven't we been moving away from proactively creating the ripple element for performance reasons and waiting for user interaction to do so?  Also applies to the usage in `setType` below.

samrichardsontylertech: I'm not even sure the ripple should be included in the component at all. It needs some more thought for sure.

src/lib/expansion-panel/expansion-panel.ts
samrichardsontylertech: I think you're right. I'll remove them.

PR #266, opened 2023-04-21
[list-item][select] allow list-item/option text to wrap

### LINKED ISSUE #207: [list-item] wrap attribute/property not working
**Describe the bug:**
When using the 'wrap' attribute on a `<forge-list-item>` element, the text within the list item is not wrapping and is being cut off with ellipsis regardless.

**To Reproduce:**
Create a forge-list with a set width and add enough text within the <forge-list-item> to trigger overflow/wrapping

**Expected behavior:**
Text should wrap to another line

**Screenshots:**
image

 - Forge version: 2.x
 - Browser: All
 - Platform: All
 - OS:  ALl
 
 
 ** CODE SNIPPET TO REPRODUCE **
```html
    <div style="width: 320px">
      <forge-list>
        <forge-list-item two-line wrap>
          <span slot="title">List Item One</span>
          <div slot="subtitle">Lorem ipsum dolor sit amet consectetur adipisicing elit. Natus illo laborum ipsum maiores sed maxime odio cupiditate ea autem exercitationem? Rem, culpa exercitationem. Doloribus laborum officiis voluptates quo error ipsam!</div>
        </forge-list-item>
        <forge-list-item two-line>
          <span slot="title">List Item Two</span>
          <span slot="subtitle">Secondary Text</span>
        </forge-list-item>
        <forge-list-item two-line>
          <span slot="title">List Item Three</span>
          <span slot="subtitle">Secondary Text</span>
        </forge-list-item>
        <forge-list-item two-line>
          <span slot="title">List Item Four</span>
          <span slot="subtitle">Secondary Text</span>
        </forge-list-item>
      </forge-list>
    </div>
```

### LINKED ISSUE #218: [select] property to allow items to not be ellipsed
**Describe the feature:**

On mobile, when a forge select drop down has long list items, the pop up is off screen and list items are truncated with an ellipse (see ticket screenshots). RW would like these items to display in full and not be off screen. I created a temp. solution within RW but would like to see about integrating this back into Forge so others could use this format if they wanted.

The current working POC, the `when-size` is project specific and would need the forge break point solution:

```
forge-popup {
    @include when-size(small) {
        max-width: calc(100% - (16px*2));
        forge-list-item::part(text-container) {
            text-overflow: unset;
            white-space: normal;
            overflow: visible;
            line-height: 1;
        }
    }
}
```

`syncPopupWidth` is an alternative but this would force the popup to be the width of the select, the only current issue is with the mobile display of this drop down and this would cause the desktop version to be smaller in some instances.

**Additional context:**

https://tylerjira.tylertech.com/browse/RPTW-2600, contains screenshots of the issue and possible solution.

DRiFTy17:
## PR Checklist

- Docs have been added / updated: Y

## Describe the new behavior?

- [list-item] fixed the list-item component to ensure that setting the `wrap` property/attribute will allow text to wrap on slotted elements.
- [select] added new properties/attributes to the select component to allow for controlling whether the popup will allow for options to wrap or not. We may want to make this the default, but in an effort to avoid breaking changes we have made this opt-in.
- [popup] added the ability to specify if the popup should constrain itself to the width of the viewport as a `max-width`.
- [select] set the dropdown popup element to constrain to the viewport width by default. This seems like a good default to have, otherwise popups with long option text could expand off screen. Devs will need to opt-in to the new option wrap functionality if they want their options to wrap.

## Additional information
Fixes #207 and closes #218

### DIFF
--- a/src/lib/popup/popup.scss
+++ b/src/lib/popup/popup.scss
@@ -9,3 +9,7 @@
 :host([hidden]) {
   display: none;
 }
+
+:host([constrain-viewport-width]) {
+  max-width: 100vw;
+}
--- a/src/dev/pages/list/list.ejs
+++ b/src/dev/pages/list/list.ejs
@@ -11,6 +11,51 @@
     </div>
   </div>
 
+  <!-- Single-line long text -->
+  <div>
+    <h3 class="forge-typography--subtitle1">Single-Line (long text)</h3>
+    <div class="list-demo">
+      <forge-list>
+        <forge-list-item>Lorem ipsum dolor sit amet consectetur adipisicing elit. Dignissimos, libero nemo. Similique suscipit ipsam ab veritatis excepturi repellendus ipsa labore! Accusantium, sint ex. Illum architecto rerum, voluptates consectetur cum quaerat.</forge-list-item>
+        <forge-list-item>Lorem ipsum dolor sit amet consectetur adipisicing elit. Dignissimos, libero nemo. Similique suscipit ipsam ab veritatis excepturi repellendus ipsa labore! Accusantium, sint ex. Illum architecto rerum, voluptates consectetur cum quaerat.</forge-list-item>
+        <forge-list-item>Lorem ipsum dolor sit amet consectetur adipisicing elit. Dignissimos, libero nemo. Similique suscipit ipsam ab veritatis excepturi repellendus ipsa labore! Accusantium, sint ex. Illum architecto rerum, voluptates consectetur cum quaerat.</forge-list-item>
+      </forge-list>
+    </div>
+  </div>
+
+  <!-- Single-line long text w/wrap -->
+  <div>
+    <h3 class="forge-typography--subtitle1">Single-Line (long text w/wrap)</h3>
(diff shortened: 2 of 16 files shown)


src/lib/popup/popup.scss
```diff
@@ -9,3 +9,7 @@
 :host([hidden]) {
   display: none;
 }
+
+:host([constrain-viewport-width]) {
+  max-width: 100vw;
+}
```
DRiFTy17, line 15: @jake-crawford-tyler I know your original issue essentially had this specified inside of a media query. I felt that utilizing an attribute for this would be ideal since the select component could set it by default going forward. So basically regardless of viewport size, the `max-width` will be set going forward, and you can now opt-in to allowing your options to wrap. No need to worry about media queries since it will render the same on both desktop and mobile.

Any concerns here from you given your original use case?

jake-crawford-tyler: This sounds ideal actually. I only set the media query as to not collide with any desired look beyond that. Two thumbs up for me.

jake-crawford-tyler (approved):

PR #274, opened 2023-04-27
fix(select): fixed a bug where the option group `builder` property was not being called

### LINKED ISSUE #260: [select] option group `builder` not being called
**Describe the bug:**
The `builder` callback on the `ISelectOptionGroup` objects when applied to the `<forge-select>` is not being called.

The issue is located here:
https://github.com/tyler-technologies-oss/forge/blob/main/src/lib/select/core/base-select-adapter.ts#L228

Another issue is that you can't bind a `builder` property to the `<forge-option-group>` elements because it doesn't exist on the public API and therefore you are required to set the options via the `options` property on the `<forge-select>`. We need to make sure that the `<forge-option-group>` element implements the `ISelectOptionGroup` interface to ensure the same properties can be used declaratively.

**To Reproduce:**
View the StackBlitz example here:
https://stackblitz.com/edit/forge-select-with-groups

**Expected behavior:**
The `builder` callback should be executed when set imperatively via the `options` property on the `<forge-select>`, or when used declaratively via the `<forge-option-group>` element (which will be generated if using the imperative approach anyway). 

 - Forge version: ^2.00
 - Browser: All
 - Platform: All
 - OS: All

**Additional context:**
We should validate all other properties are usable, updated any unit tests, and that the `<forge-option>` element operates as expected as well.

DRiFTy17:
## PR Checklist

## Describe the new behavior?
The `builder` property on `<forge-option-group>` elements (among other properties) will now correctly propagate to the list-dropdown.

## Additional information
Fixes #260

### DIFF
--- a/src/lib/select/core/base-select-adapter.ts
+++ b/src/lib/select/core/base-select-adapter.ts
@@ -1,4 +1,4 @@
-import { isDefined, removeElement } from '@tylertech/forge-core';
+import { isDefined, isFunction, removeElement } from '@tylertech/forge-core';
 import { BaseAdapter, IBaseAdapter } from '../../core/base/base-adapter';
 import { IBaseSelectComponent } from './base-select';
 import { ListDropdown, IListDropdown } from '../../list-dropdown';
@@ -8,7 +8,6 @@ import { IOptionGroupComponent, OPTION_GROUP_CONSTANTS } from '../option-group';
 import { ISelectOption, ISelectOptionGroup, SelectOptionListenerDestructor } from './base-select-constants';
 import { isOptionGroupObject } from './select-utils';
 import { IPopupComponent, POPUP_CONSTANTS } from '../../popup';
-import { assignMatchingProperties } from '../../core/utils/utils';
 
 export interface IBaseSelectAdapter extends IBaseAdapter {
   initializeAccessibility(): void;
@@ -64,7 +63,11 @@ export abstract class BaseSelectAdapter extends BaseAdapter<IBaseSelectComponent
       return optionGroupElements.map(optionGroupElement => {
         const optionElements = Array.from(optionGroupElement.querySelectorAll(OPTION_CONSTANTS.elementName)) as IOptionComponent[];
         const options = this._createOptionsFromElements(optionElements);
-        return { text: optionGroupElement.label, options } as ISelectOptionGroup;
+        return {
+          text: optionGroupElement.label,
+          builder: optionGroupElement.builder,
+          options
+        } as ISelectOptionGroup;
       });
     } else {
       const optionElements = Array.from(this._component.querySelectorAll(OPTION_CONSTANTS.elementName)) as IOptionComponent[];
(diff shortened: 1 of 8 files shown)


src/lib/select/core/base-select-adapter.ts
```diff
@@ -64,7 +63,11 @@ export abstract class BaseSelectAdapter extends BaseAdapter<IBaseSelectComponent
       return optionGroupElements.map(optionGroupElement => {
         const optionElements = Array.from(optionGroupElement.querySelectorAll(OPTION_CONSTANTS.elementName)) as IOptionComponent[];
         const options = this._createOptionsFromElements(optionElements);
-        return { text: optionGroupElement.label, options } as ISelectOptionGroup;
+        return {
+          text: optionGroupElement.label,
+          builder: optionGroupElement.builder,
```
DRiFTy17, line 68: The `builder` property missing from here was the main problem.

src/lib/select/core/base-select-adapter.ts
```diff
@@ -225,14 +228,13 @@ export abstract class BaseSelectAdapter extends BaseAdapter<IBaseSelectComponent
 
   private _createOptionGroupElement(group: ISelectOptionGroup): HTMLElement {
     const optionGroupElement = document.createElement('forge-option-group');
-    assignMatchingProperties(group, optionGroupElement);
-    optionGroupElement.label = group.text ?? '';
+    Object.assign(optionGroupElement, group);
```
DRiFTy17, line 231: We will now just blindly copy everything from the config objects for both options and option groups. The previous strategy of only applying "matching properties" was short sighted.

src/lib/select/core/select-utils.ts
```diff
@@ -15,7 +15,7 @@ export function isSelectOptionType(options: ISelectOption[] | ISelectOptionGroup
 }
 
 export function isOptionGroupObject(o: ISelectOption | ISelectOptionGroup): o is ISelectOptionGroup {
-  return isDefined(o) && isObject(o) && o.hasOwnProperty('options') && o.hasOwnProperty('text');
+  return isDefined(o) && isObject(o) && o.hasOwnProperty('options');
```
DRiFTy17, line 18: We do not need to ensure that the `text` property is provided since the existing of the `options` property is all we actually care about to determine if this is an option group.

src/lib/select/option-group/option-group.ts
```diff
@@ -51,4 +53,18 @@ export class OptionGroupComponent extends BaseComponent implements IOptionGroupC
...
+  /** The builder function for the group content. */
+  public declare builder: ListDropdownOptionGroupBuilder;
+
+  /** The text content for the group. */
+  public get text(): string {
+    return this._label;
+  }
+  public set text(value: string) {
+    this.label = value;
+  }
```
DRiFTy17, line 69: These properties can be set on the `<forge-option-group>` elements instead of forcing it to be done through JavaScript only.

samrichardsontylertech (approved):

PR #280, opened 2023-04-28
Feat/stack component

nickonometry:
## PR Checklist

- Docs have been added / updated: N (This will not be added to the Forge website, Storybook docs in a separate PR after release)

## Describe the new behavior?
The forge-stack is a layout utility component that makes it easy to stack elements together vertically or horizontally and apply a gap between them. This is not meant to be an abstracted API of flex or grid, it's just an easier way to provide vertical and horizontal rhythm between elements from the HTML

image


src/dev/pages/stack/stack.scss
```diff
@@ -0,0 +1,17 @@
...
+  height: 800px;
+  padding: 16px;
+  background: beige;
+  border-radius: 8px;
+}
+
+.box {
+    padding: 36px;
+    border: 4px dashed darkgray;
+    background-color: bisque;
```
DRiFTy17, line 11: Super minor, but if you wanted to make this easier on the eyes in dark mode you could handle swapping these hard coded colors around depending on theme.

src/dev/pages/stack/stack.ts
```diff
@@ -0,0 +1,54 @@
...
+import '@tylertech/forge/stack';
+
+import '$src/shared';
+import type { ISwitchComponent } from '@tylertech/forge/switch';
+import { ISelectComponent, IStackComponent, ITextFieldComponent } from '@tylertech/forge';
+
+
+const inlineToggle = document.querySelector('#inline-switch') as ISwitchComponent;
+const wrapToggle = document.querySelector('#wrap-switch') as ISwitchComponent;
+wrapToggle.disabled = true;
```
DRiFTy17, line 15: FYI, you can set `disabled: true` on the option configuration to default it to disabled.

src/lib/stack/stack-foundation.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  /** Controls the gap amount between the children in pixels */
+  public get gap(): number {
+    return this._gap;
+  }
+  public set gap(value: number) {
+    if (this._gap !== value) {
+      this._gap = value;
+      this._adapter.setGap(this._gap);
+      this._adapter.setHostAttribute(STACK_CONSTANTS.attributes.GAP, String(this._gap));
```
DRiFTy17, line 66: It may be better to accept both a `number` and `string` here for non-pixel values as well as supplying the gap as a string with a `px` value for convenience? Setting to `16px` causes an exception currently, and there may be cases where supplying a percentage or some other non-discrete values.

src/lib/stack/stack.ts
```diff
@@ -0,0 +1,94 @@
...
+import template from './stack.html';
+import styles from './stack.scss';
+import { BaseComponent } from '../core';
+
+export interface IStackComponent extends ICustomElement {
+  inline: boolean;
+  wrap: boolean;
+  stretch: boolean;
+  gap: number;
+  align: StackAlignMode;
```
DRiFTy17, line 15: We may want to rename this to `alignment` since HTML already has an `align` attribute and it automatically applies `text-align` when this attribute is set.

src/lib/stack/stack.ts
```diff
@@ -0,0 +1,90 @@
...
+declare global {
+  interface HTMLElementTagNameMap {
+    'forge-stack': IStackComponent;
+  }
+}
+
+@CustomElement({
+  name: STACK_CONSTANTS.elementName
+})
+export class StackComponent extends BaseComponent implements IStackComponent {
```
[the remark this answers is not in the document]
nickonometry: Will do

src/lib/stack/_mixins.scss
```diff
@@ -0,0 +1,44 @@
...
+  display: flex;
+  flex-direction: row;
+  --forge-stack-height: auto;
+}
+
+@mixin wrap() {
+  flex-wrap: wrap;
+}
+
+@mixin stretch() {
```
samrichardsontylertech, line 28: I'm curious if it would be possible to enable "tracks" to stretch instead of individual items, so items in the bottom row are the same size as the ones above instead of filling the entire width. That would probably mean switching to `grid`... is that too big an ask?

samrichardsontylertech: That in addition to the way it's handled now. just as an extra option

src/dev/pages/stack/stack.ejs
```diff
@@ -0,0 +1,119 @@
...
+          <forge-text-field>
+            <input type="text" id="input-text-01" />
+            <label for="input-text-01" slot="label">Text field</label>
+          </forge-text-field>
+        </forge-stack>
+      </form>
+    </forge-card>
+
+
+    <forge-card outlined style="--forge-card-padding: 0;">
```
samrichardsontylertech, line 52: I'm worried that using this as-is for complex layouts will discourage responsive design — each row in this example will keep its number of columns even in a really small container. Ideally the items should be allowed to wrap to a new line when the stack narrows.

I think this should introduce something like a `data-stack-break` attribute to tell the component where to start new lines as the container size changes. Basically like a `grid-column-start: 1` style declaration. It might look like this:
```
<forge-stack inline wrap stretch>
  <div>Full width</div>
  <div data-stack-break>Half width</div>
  <div>Half width</div>
  <div data-stack-break>Third width</div>
  <div>Third width</div>
  <div>Third width</div>
</forge-stack>
```
As the container shrinks it would gradually move to a single column layout, made possible because the inner stacks have been flattened into the base stack.

samrichardsontylertech:
Have you tested the accessibility implications of this? I expect it will be used to style lists pretty often so being able to do something like this will be important:
```
<forge-stack aria-role="list">
  <div aria-role="listitem></div>
  <div aria-role="listitem></div>
  <div aria-role="listitem></div>
</forge-stack>
```
I'm not sure how generic elements between the list and listitems (in or out of the shadow DOM) affect their relationship.

src/dev/pages/stack/stack.ejs
samrichardsontylertech: This is another change that might require `grid`

samrichardsontylertech (changes requested):
This largely looks good as is, but I want to make sure this can be intuitively accessible and give you a chance to consider my suggestions before approving it.

src/lib/stack/_mixins.scss
```diff
@@ -0,0 +1,44 @@
...
+}
+
+@mixin base() {
+  display: flex;
+  flex-direction: column;
+  width: 100%;
+}
+
+@mixin inline() {
+  display: flex;
```
DRiFTy17, line 19: This is repetitive since the `base` mixin already sets it.

src/lib/stack/_mixins.scss
```diff
@@ -0,0 +1,44 @@
...
+
+@mixin core-styles() {
+  .forge-stack {
+    @include base();
+    @include theme.css-custom-property(gap, --forge-stack-gap, variables.$gap);
+    @include theme.css-custom-property(height, --forge-stack-height, 100%);
+  }
+}
+
+@mixin base() {
```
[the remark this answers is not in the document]
DRiFTy17: I think since the goal of this component to is to only lay out items with a gap between them in specified orientation that it will need either `flex` or `grid`?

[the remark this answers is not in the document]
DRiFTy17: I think since the goal of this component to is to only lay out items with a gap between them in specified orientation that it will need either `flex` or `grid`?

src/lib/stack/_mixins.scss
DRiFTy17: I believe the intention is to keep this one dimensional. Nick and I discussed this potentially using `grid` to allow for multi dimensional layouts but I think it goes beyond the scope of what this basic layout component is intended for. Whether we use `grid` or `flex` probably doesn't matter much unless we get better responsive support with `grid`?

src/dev/pages/stack/stack.ejs
DRiFTy17: Being able to control where/when it wraps is a good point here that may make a better case for `grid` instead. Right now everything will collapse down into a single column, but `flex` doesn't provide an easy way like `grid` does.

I think it worth noting that we should also abstract any concept of `flex` or `grid` from the API so we can make modifying the internals easier without technologies-specific naming.

src/lib/stack/_mixins.scss
samrichardsontylertech: `grid` is definitely capable of more complex responsiveness but you lose some perks like independently sized items. There's a tradeoff for sure. Some of the downsides to stretching everything could be mitigated by creatively setting the width of items.

src/lib/stack/_mixins.scss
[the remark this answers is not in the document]
DRiFTy17: Ah, yep, definitely agree!

src/lib/stack/stack-constants.ts
```diff
@@ -0,0 +1,32 @@
...
+
+const classes = {
+  DEFAULT: 'forge-stack'
+};
+
+const attributes = {
+  INLINE: 'inline',
+  WRAP: 'wrap',
+  STRETCH: 'stretch',
+  GAP: 'gap',
```
samrichardsontylertech, line 13: Should this use the same terminology as CSS or would it be more appropriate to call this attribute `spacing` or something like that?

DRiFTy17: Yeah I'm on board with sticking with `inline`

Lillious (dismissed):

samrichardsontylertech (dismissed):

DRiFTy17 (approved):

PR #285, opened 2023-05-01
perf: checkbox, radio, and switch will now instantiate ripple upon first user interaction

### LINKED ISSUE #242: [radio] ripple is misaligned when placed inside of a `<forge-dialog>`
**Describe the bug:**
The ripple within a `<forge-radio>` element is not aligned properly when placed in a `<forge-dialog>` because of the `transition` that is applied to the `transform: scale()` on the dialog surface. This causes the initial ripple calculation to be incorrect until the user interacts with the element.

The dialog currently tries to handle this situation for common elements as can be seen here:
https://github.com/tyler-technologies-oss/forge/blob/04741554822b7b4b5831d5146cf1ed4f4ed225a0/src/lib/dialog/dialog-adapter.ts#L160-L164

**To Reproduce:**
Place the following HTML inside of a `<forge-dialog>`:
```html
<div role="radiogroup">
  <forge-radio>
    <input type="radio" name="test-1" id="test-1" checked />
    <label for="test-1">One</label>
  </forge-radio>
  <forge-radio>
    <input type="radio" name="test-2" id="test-2" />
    <label for="test-2">Two</label>
  </forge-radio>
  <forge-radio>
    <input type="radio" name="test-3" id="test-3" />
    <label for="test-3">Three</label>
  </forge-radio>
</div>
```

Open the dialog, and set focus to the first radio button. Observe the misalignment shown below.

**Expected behavior:**
The ripple should be aligned properly regardless of where it's used.

**Screenshots:**
image

 - Forge version: ^2.0.0
 - Browser: All
 - Platform: All
 - OS: All

**Additional context:**
We [recently](https://github.com/tyler-technologies-oss/forge/blob/04741554822b7b4b5831d5146cf1ed4f4ed225a0/src/lib/icon-button/icon-button.ts#L164-L173) adjusted the icon-button component to use JIT-style ripple initialization logic to avoid instantiating the ripple until user interaction occurs. That would be a better way to resolve this problem by migrating that same solution to other components that utilize a ripple.

### LINKED ISSUE #256: [checkbox] performance issue when toggling checked state
**Describe the bug:**
When toggling the `checked` state of the `<input>` element within a `<forge-checkbox>` if there are many (at least 100) instances in the DOM at the same time, there is a noticeable render lag occurring if toggling all instances simultaneously.

This appears to be related to the ripple since it is re-running its layout logic every time the checked state changes. This shouldn't need to occur.

We need to also implement a similar JIT-style initialization of the ripple as was done recently with the icon button in #110.

**To Reproduce:**
1. Create a layout where there are 100+ elements on the page that each have a `<forge-checkbox>`
2. Update the checked state via JavaScript on all 100+ `<forge-checkbox>` elements
3. Observe a render-blocking lag delay

**Expected behavior:**
The checked state should updated immediately.

 - Forge version: ^2.0.0
 - Browser: All
 - Platform: All
 - OS: All

**Additional context:**
This was found in a `<table>` implementation where there were 100 rows, and after further debugging it was not due to re-rendering of the checkboxes but only caused when toggling the checked state of each checkbox at the same time.

We will also want to test and inspect the radio and switch since those components operate in a similar fashion.

DRiFTy17:
## PR Checklist

## Describe the new behavior?
Several related changes were made regarding performance concerns with how/when the MDC ripple is instantiated. Overall the goal is to delay ripple instantiation until first user interaction to avoid the overhead the ripple causes.

Changes made:
- Delay ripple instantiation until first user interaction on checkbox, radio, and switch
- Remove unnecessary `layout()` method calls in the checkbox when toggling checked state
  - Tested with 1000 checkboxes on a page and noticed very little lag after this change.
- Remove legacy code in the dialog that would attempt to call the `layout()` method on checkbox, radio, and switch to "fix" incorrect ripple alignment after opening due to `transform: scale()`.
  - This is no longer necessary given that the ripple won't be immediately instantiated.
- Updated existing implementations of this pattern on button, icon-button, and list-item to fix some syntax inconsistencies, and reduce unsafe access to protected members of `MDCRipple`.

## Additional information
Fixes #256 
Fixes #242

### DIFF
--- a/src/lib/button/button.ts
+++ b/src/lib/button/button.ts
@@ -130,19 +130,18 @@ export class ButtonComponent extends BaseComponent implements IButtonComponent {
   private async _deferRippleInitialization(): Promise<void> {
     const type = await userInteractionListener(this._buttonElement);
     if (!this._rippleInstance) {
-      this._initRipple();
+      this._rippleInstance = this._createRipple();
       if (type === 'focusin') {
-        // eslint-disable-next-line @typescript-eslint/dot-notation
-        (this._rippleInstance as ForgeRipple)['foundation'].handleFocus();
+        this._rippleInstance.handleFocus();
       }
     }
   }
 
-  private _initRipple(): void {
+  private _createRipple(): ForgeRipple {
     if (this._rippleInstance) {
       this._rippleInstance.destroy();
     }
-    this._rippleInstance = new ForgeRipple(this._buttonElement);
+    return new ForgeRipple(this._buttonElement);
   }
 
   private _initializeButtonChildren(): void {
--- a/src/lib/checkbox/checkbox-adapter.ts
+++ b/src/lib/checkbox/checkbox-adapter.ts
@@ -1,4 +1,5 @@
 import { addClass, getShadowElement, removeClass } from '@tylertech/forge-core';
+import { userInteractionListener } from '../core';
 import { BaseAdapter, IBaseAdapter } from '../core/base/base-adapter';
 import { ForgeRipple, ForgeRippleAdapter, ForgeRippleCapableSurface, ForgeRippleFoundation } from '../ripple';
 import { ICheckboxComponent } from './checkbox';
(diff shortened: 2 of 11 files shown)


MikeMatusz (approved):
Nice how this actually simplified a number of components.

PR #290, opened 2023-05-03
fix(chip-field): fixed a bug where addon-end slot was not expanding full height as the field grew

### LINKED ISSUE #216: [chip-field] addon-end slot not expanding/aligning properly
**Describe the bug:**
The chip-field differs from the text-field when placing content within the `addon-end` slot that causes the field to grow in height.

An example of the issue can be seen here from this reproduction:
image

**To Reproduce:**
1. Add chips to the chip-field that cause the field to grow in height

**Expected behavior:**
The chip-field should properly expand the divider and vertically center the content within the `addon-end` slot. The text-field itself already handles this properly, so we have a bug in the shared styles.

 - Forge version: `^2.0.0`
 - Browser: All
 - Platform: All
 - OS: All

DRiFTy17:
## PR Checklist

## Describe the new behavior?
The `addon-end` container will now properly grow with the field height.

image

Some other internal styles were not being applied properly due to incorrect selector nesting, such as height CSS custom properties, and those will now work properly as well as a side effect of this.

## Additional information
Fixes #216

### DIFF
--- a/src/dev/pages/chip-field/chip-field.ejs
+++ b/src/dev/pages/chip-field/chip-field.ejs
@@ -1,6 +1,6 @@
 <div class="vert">
   <forge-chip-field id="demo-simple-chip-field">
-    <label slot="label" for="demo-chip-field-input">Choose state</label>
+    <label slot="label" for="demo-chip-field-input">Test label</label>
     <input type="text" id="demo-chip-field-input">
   </forge-chip-field>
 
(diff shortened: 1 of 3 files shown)


MikeMatusz:
Should a similar change be made to handle a `textarea` inside a `text-field`, or would you rather address that separately?
image

DRiFTy17:
> Should a similar change be made to handle a `textarea` inside a `text-field`, or would you rather address that separately?

Good catch! I'll create a separate issue for that one.

MikeMatusz (approved):

PR #293, opened 2023-05-04
fix(menu): fixed toggle element initialization

### LINKED ISSUE #249: [menu] Menu doesn't close when button isn't in the DOM yet at time of initialization
**Describe the bug:**
When `<forge-menu>` is used with a button that is not yet in the DOM at time of initialization, the menu is unable to be closed by clicking off. This could be a problem, especially such as in frameworks like Angular with components that are button wrappers. In our instance, we have a button component and the button element is ngIf'd out of the DOM until a condition is met (it's bound to an | async pipe). It appears that if there is no match for the menu selector, it defaults to just assuming the first child element is the button. This is not valid, as the component only behaves properly if it's in fact a button, and this could result in a targetElement that is not a button (exactly what's happening in our scenario).

Specifically, this line of code is problematic, and should likely wait until it finds a selector match before setting the targetElement

https://github.com/tyler-technologies-oss/forge/blob/04741554822b7b4b5831d5146cf1ed4f4ed225a0/src/lib/menu/menu-adapter.ts#L54

**To Reproduce:**
1. Use a <forge-menu> with a nested <div> (or some non-button element inside)
2. Add a button to the <div> after forge-menu has been initialized (or if using angular, can simply ngIf this button into the dom after initiailization)
3. Notice the button works to open the menu, but you cannot click off to close.

**Expected behavior:**
Clicking off of the forge-menu should always close the forge-menu

 - Forge version: v2.13.1
 - Browser: All
 - Platform: All
 - OS: All

DRiFTy17:
## PR Checklist

## Describe the new behavior?
- Updated selector list for detecting toggle elements
- Updated the `connectedCallback()` to wait for a specific toggle element from the selector list before initializing
- Use the `focusout` event instead of `blur` to allow for detecting focus being lost from any child element (nested or not)

## Additional information
Fixes #249

### DIFF
--- a/src/lib/menu/menu-constants.ts
+++ b/src/lib/menu/menu-constants.ts
@@ -9,7 +9,7 @@ const classes = {
 };
 
 const selectors = {
-  TOGGLE: `.${COMPONENT_NAME_PREFIX}menu__toggle, [${elementName}-toggle], button, [type=button]`,
+  TOGGLE: `.${elementName}__toggle,[${elementName}-toggle],button,[type=button],[role=button],a,[tabindex]:not([tabindex^="-"])`,
   MENU_LIST: 'forge-list'
 };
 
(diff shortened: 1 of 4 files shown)


DRiFTy17:
This change makes the initialization of this component slightly more strict because we require an element from the selector list. This technically could cause a breaking change, but it's very unlikely.

We could revert back to the original init logic to detect when there is just any child element provided, and rely solely on the `focusout` event if that seems more safe? Curious to hear thoughts from others.

PR #298, opened 2023-05-05
fix(busy-indicator): fixed various accessibility issues

### LINKED ISSUE #255: [busy-indicator] allow setting aria-label on internal progress indicator
**Describe the bug:**
An `aria-label` is required on elements with the `progressbar` role, but the busy indicator component does not currently allow this to be set on its internal progress indicator.

**To Reproduce:**
1. Open a busy indicator with a progress indicator
2. Run an audit tool or inspect the progress indicator to see that it does not have a label

**Expected behavior:**
It should be possible to add the required `aria-label` to any instance of a progress indicator. Either the `ariaLabel` property should be included on the busy indicator as it is for the progress indicator components, or the `titleText` property should also set the `aria-label`.

 - Forge version: 2.13.2
 - Browser: All
 - Platform: All
 - OS: All

DRiFTy17:
## PR Checklist

## Describe the new behavior?
- Use `<div>` instead of `<aside>` since we set `role` anyway
- Replace `<h3>` with `<h1>`
- Hide `<forge-circular-progress>` and `<forge-linear-progress>` from the AOM using `aria-hidden="true"` because these elements are intended to decorative.
- Use an ARIA live region on the message text and convert to a `<p>` tag.

## Additional information
Fixes #255

### DIFF
--- a/src/lib/busy-indicator/busy-indicator.html
+++ b/src/lib/busy-indicator/busy-indicator.html
@@ -1,10 +1,10 @@
 <template>
   <forge-backdrop part="backdrop" fixed></forge-backdrop>
-  <div class="forge-busy-indicator__surface" role="alertdialog" aria-modal="true" part="surface" tabindex="0">
-    <h1 class="forge-busy-indicator__title" part="title">Loading</h1>
+  <div class="forge-busy-indicator__surface" role="alert" aria-labelledby="title" aria-describedby="message" part="surface" tabindex="0">
+    <h1 id="title" class="forge-busy-indicator__title" part="title">Loading</h1>
     <div class="forge-busy-indicator__message-container" part="message-container">
       <forge-circular-progress class="forge-busy-indicator__spinner" part="spinner" aria-hidden="true"></forge-circular-progress>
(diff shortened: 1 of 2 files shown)


src/lib/busy-indicator/busy-indicator.html
```diff
@@ -1,16 +1,16 @@
 <template>
   <forge-backdrop part="backdrop" fixed></forge-backdrop>
-  <aside class="forge-busy-indicator__surface" role="alertdialog" aria-modal="true" part="surface" tabindex="0">
-    <h3 class="forge-busy-indicator__title" part="title">Loading</h3>
+  <div class="forge-busy-indicator__surface" role="alertdialog" aria-modal="true" part="surface" tabindex="0">
```
samrichardsontylertech, line 3: This should include `aria-labelledby` and `aria-describedby` attributes to conform to W3's example

https://www.w3.org/WAI/ARIA/apg/patterns/alertdialog/examples/alertdialog/

samrichardsontylertech (changes requested):
Looks better! Add accessible attributes to the alert dialog and I think it'll be good to go

PR #313, opened 2023-05-23
feat(autocomplete): add `filterFocusFirst` property

DRiFTy17:
## PR Checklist

## Describe the new behavior?
When filtering options by typing, the dropdown will focus the first option automatically. This change adds a new `filterFocusFirst` configuration property that defaults to `true` to retain the existing functionality, but allows for a developer to set it to `false` to stop the first option from being automatically focused if desired.

## Additional information
This came up due to an application needing to allow for submitting a form via the enter key (specifically when using `allowUnmatched` and composed with a chip-field), and it was confusing for users that the selection was changing because the first option was always automatically activated in the dropdown.


MikeMatusz:
In the [PSDF combobox|https://www.psdf.docs.tylerdev.io/?path=/story/components-combobox--combobox], we had the requirement that filtering and hitting enter or tab would select the active item, which is definitely more like a desktop application behavior than typical web UI, but we're definitely relying on this behavior.  As mentioned on Slack, but adding this comment for visiblity, I think this should go through Design Governance if it's not going to be optional.

MikeMatusz (approved):

Issue #314, opened 2023-05-23
[date-picker][time-picker] Behavior of Mask in Pristine Condition


jake-crawford-tyler:
**Describe the bug:**
Behavior of the mask in date-picker and time-picker while in pristine condition, `masked=true`, and `showMaskFormat=true` seems to be unexpected. While the component is pristine, the mask should only show while the user is focused. Otherwise, if the user tabs away and the component is still pristine, the mask will hide.

May loosely relate to:
https://github.com/tyler-technologies-oss/forge/issues/229

**To Reproduce:**
Steps to reproduce the behavior, same for all listed components:

1. Go to time-picker
2. `masked=true` and `showMaskFormat=true`
3. Note the Mask is already displaying, not expected
4. Click on input
5. Note the Mask is showing, expected
6. Do not enter data
7. Tab or click away from the input
8. Mask is not showing, label appears to be a placeholder, expected

**Expected behavior:**
Expected behavior, same for all listed components:

1. Go to time-picker
2. `masked=true`, and `showMaskFormat=true`
3. Mask is not showing, label appears to be a placeholder
4. Click on input
5. Mask shows on input
6. Do not enter data
7. Tab or click away from the input
8. Mask is not showing, label appears to be a placeholder

**Screenshots:**

**Please complete the following information:**
 - Forge version: [v2.14.0]
 - I have searched existing issues before creating this report? [Y]
 - Browser: [All]
 - Platform: [All]
 - OS: [All]

**Additional context:**

PR #318, opened 2023-05-25
fix(masks): time-picker, date-picker, date-range-picker > updates to mask behavior

jake-crawford-tyler:
## PR Checklist

## Describe the new behavior?
time-picker, date-picker, and date-range-picker will now only display masks when they are focused. Also masks will hide on blur when no value is populated.

## Additional information
Closes: https://github.com/tyler-technologies-oss/forge/issues/314

### DIFF
--- a/src/lib/time-picker/time-picker-foundation.ts
+++ b/src/lib/time-picker/time-picker-foundation.ts
@@ -126,6 +126,7 @@ export class TimePickerFoundation implements ITimePickerFoundation {
     this._applyAllowInput();
     this._applyAllowDropdown();
     this._applyDisabled();
+    this._applyMask();
 
     this._isInitialized = true;
   }
@@ -289,14 +290,19 @@ export class TimePickerFoundation implements ITimePickerFoundation {
     if (this._allowInput) {
       this._adapter.selectInputText();
     }
-    this._applyMask();
+
+    if (this.masked && this._showMaskFormat) {
+      this._applyMask();
+    }
   }
 
   private _onInputBlur(evt: Event): void {
-    if(this.masked && this.showMaskFormat) {
-      this._adapter.destroyMask();
+    if (this.masked && this._showMaskFormat) {
+      this._applyMask();
     }
+
     this._formatInputValue();
+
     if (this._open && !this._adapter.isInputFocused()) {
       this._closeDropdown(true);
     }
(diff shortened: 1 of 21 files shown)


DRiFTy17 (commented):
This looks good to me from a functional perspective. It gets me wondering though, do we want to just always leave the mask uninitialized until the field is first focused instead? The input masking functionality is only needed while the user is typing into the field. I like that we no longer have the mask overhead while the field is resting since they are not needed at that time...

A couple of minor issues relating to the floating label jumped out at me as well:
- The label doesn't animate to its floating position when the mask is enabled on focus
- The label jumps slightly when tabbing between the inputs in the date-range-picker

src/lib/time-picker/time-picker-foundation.ts
```diff
@@ -769,9 +772,11 @@ export class TimePickerFoundation implements ITimePickerFoundation {
     if (this._allowSeconds !== value) {
       this._allowSeconds = !!value;
       this._applyAllowSeconds();
-      if (this._isInitialized) {
+      if (this._isInitialized && !this._showMaskFormat) {
```
madhater, line 775: this seems backwards, if not showMaskFormat then apply the mask?

src/test/spec/date-range-picker/date-range-picker.spec.ts
```diff
@@ -512,6 +512,7 @@ describe('DateRangePickerComponent', function(this: ITestContext) {
       const date = '05/04/2020';
       const expectedDate = new Date(date);
       const inputElement = getToElement(this.context.component);
+      inputElement.focus();
```
jake-crawford-tyler, line 515: Honestly, this one was a bit confusing as the "from" unit test did not require a focus. I am open to suggestions on if this is an incorrect test.

src/lib/time-picker/time-picker-foundation.ts
jake-crawford-tyler: After re-evaluating. I still think this is correct. If this condition was not here, then applyMask function will make the mask show if showMaskFormat is true. When testing in the UI, turning on and off allowSeconds and use24HourTime will cause the mask to show when it should not. Still open to discussion on this one @madhater and @DRiFTy17

jake-crawford-tyler:
I will need assistance tracking down the issue with the animations. How should I proceed from here? @DRiFTy17

DRiFTy17:
> I will need assistance tracking down the issue with the animations. How should I proceed from here? @DRiFTy17

@jake-crawford-tyler I started looking into this earlier today, but I've been hopping back and forth between a number of different tasks all day... I was able to fix the issue regarding the floating label animation when focusing the field with the mask format visible though.

I still need to track down the issue in the date-range-picker when tabbing between the inputs, and then finish reviewing the latest changes.

Sit tight for a bit and I'll update your PR with some changes, and give it all another review.

jake-crawford-tyler:
Perfect, thank you @DRiFTy17 , I appreciate the help!

DRiFTy17 (approved):

PR #323, opened 2023-06-08
feat(circular-progress): refactor to remove MDC

DRiFTy17:
## PR Checklist

## Describe the new behavior?
Refactor the `<forge-circular-progress>` component to remove legacy MDC JavaScript/CSS. This also includes an initial implementation for how a basic "style layer" separation might look (subject to change further).

The `open` property/attribute was removed from the component as this should not be a responsibility of the component itself going forward.

Additionally a new "track" background element was added that shows the circular track at all times by default both determinate and indeterminate state:
image


src/lib/circular-progress/circular-progress-adapter.ts
```diff
@@ -1,56 +1,52 @@
...
-  }
-
   public setDeterminate(value: boolean): void {
-    if (this._mdcCircularProgress) {
-      this._mdcCircularProgress.determinate = value;
+    this._tryResetTemplate();
+    this._rootElement.classList.toggle(CIRCULAR_PROGRESS_CONSTANTS.classes.INDETERMINATE, !value);
+    
+    if (value) {
+      this._rootElement.insertAdjacentHTML('beforeend', determinateTemplate);
```
MikeMatusz, line 30: Should we also remove the opposite template if already inserted, e.g. if `determinate` were toggled while already running.  Probably not the most typical usage pattern, but if I'm reading this right, seems you could end up with multiple spinners.

src/lib/circular-progress/circular-progress.ts
```diff
@@ -21,20 +20,41 @@ declare global {
...
+ * 
+ * @slot - The is the default/unnamed slot. Renders content at the center of the progress indicator.
+ * 
+ * @cssproperty --forge-circular-progress-size - The height and width of the indicator container.
+ * @cssproperty --forge-circular-progress-padding - The padding inside the bounding box of the container.
+ * @cssproperty --forge-circular-progress-indicator-width - The active indicator width.
+ * @cssproperty --forge-circular-progress-indicator-color - The active indicator color.
+ * @cssproperty --forge-circular-progress-indicator-background-color - The indicator background color.
+ * 
+ * @csspart progressbar - Styles the progress bar container element
```
MikeMatusz, line 47: Nice to see all these custom manifest attributes being specified!  Definitely would be nice to get the corresponding portions of the docs generated.

src/lib/core/style-layer/circular-progress/_core.scss
```diff
@@ -0,0 +1,126 @@
...
+
+///
+/// Initializes private configuration state CSS variables
+/// 
+@mixin configuration {
+  --_size: var(--#{variables.$prefix}-size, #{variables.$size});
+  --_spacing: var(--#{variables.$prefix}-padding, #{variables.$spacing});
+  --_track-width: var(--#{variables.$prefix}-indicator-width, #{variables.$track-width});
+  --_track-color: var(--#{variables.$prefix}-indicator-color, var(--forge-theme-tertiary, #{theme-values.$tertiary}));
+  --_track-background: var(--#{variables.$prefix}-indicator-background-color, var(--mdc-theme-text-disabled-on-background, #{theme-values.$text-disabled-on-background}));
```
MikeMatusz, line 17: This might be a separate discussion, but is the goal to eventually have all `--forge-` variables and no `--mdc-`, or are there still some central theme ones we'll always use?

src/lib/circular-progress/circular-progress-adapter.ts
DRiFTy17: I _think_ this is working as expected currently because it calls `this._tryResetTemplate()` which will remove either of the existing templates (indeterminate or determinate). I chose to keep it simple that way because the likelihood of someone toggling between determinate and indeterminate dynamically is probably going to be pretty rare so performance wasn't a concern of mine...

Definitely open to any suggestions you may have though!

src/lib/circular-progress/circular-progress.ts
DRiFTy17: I plan to do a POC of generating the docs from the manifest pretty soon. I'll follow up with you on any findings when I get there.

PR #326, opened 2023-06-14
feat(button-area): create button area component

samrichardsontylertech:
## PR Checklist

## Describe the new behavior?
This introduces a new button area component that listens for clicks while using a visually hidden button element to accessibly enable complex and interactive nested content in clickable areas.

### DIFF
--- a/src/dev/pages/button-area/button-area.ejs
+++ b/src/dev/pages/button-area/button-area.ejs
@@ -1,0 +1,20 @@
+<forge-card outlined="true">
+  <forge-button-area id="button-area">
+    <button slot="button">Heading</button>
+    <div class="flex-wrapper">
+      <div>
+        <div class="forge-typography--headline5">Heading</div>
+        <div>Content</div>
+      </div>
+      <forge-icon-button data-forge-ignore>
+        <button>
+          <forge-icon name="favorite"></forge-icon>
+        </button>
+        <forge-tooltip>Like</forge-tooltip>
+      </forge-icon-button>
+      <forge-icon name="chevron_right"></forge-icon>
+    </div>
+  </forge-button-area>
+</forge-card>
+
+<script type="module" src="button-area.ts"></script>
--- a/src/dev/pages/button-area/button-area.ejs
+++ b/src/dev/pages/button-area/button-area.ejs
@@ -1,4 +1,4 @@
-<forge-card outlined="true">
+<forge-card outlined>
   <forge-button-area id="button-area">
     <button slot="button">Heading</button>
     <div class="flex-wrapper">
--- a/src/dev/pages/button-area/button-area.ejs
+++ b/src/dev/pages/button-area/button-area.ejs
@@ -1,9 +1,9 @@
 <forge-card outlined>
   <forge-button-area id="button-area">
-    <button slot="button">Heading</button>
+    <button slot="button" aria-labelledby="button-area-heading"></button>
     <div class="flex-wrapper">
       <div>
-        <div class="forge-typography--headline5">Heading</div>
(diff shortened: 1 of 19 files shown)


src/lib/index.ts
```diff
@@ -139,6 +139,7 @@ export * from './utils';
 export * from './view-switcher';
 export * from './icon';
 export * from './stack';
+export * from './button-area';
```
DRiFTy17, line 142: We also want to import the `ButtonAreaComponent` class and add it to the `CUSTOM_ELEMENTS` array below.

src/lib/button-area/button-area.ts
```diff
@@ -0,0 +1,60 @@
...
+  disabled: boolean;
+}
+
+declare global {
+  interface HTMLElementTagNameMap {
+    'forge-button-area': IButtonAreaComponent;
+  }
+}
+
+@CustomElement({
```
DRiFTy17, line 20: Just a reminder to add the comment here for the custom elements manifest.

src/dev/pages/button-area/button-area.ejs
```diff
@@ -0,0 +1,20 @@
+<forge-card outlined="true">
```
DRiFTy17, line 1: Super minor, but can we remove the `="true"` part of this attribute? I know it doesn't matter, but technically with boolean attribute just the _existence_ matters so I have been trying to remove instances of this to avoid setting the example.

src/lib/button-area/button-area-adapter.ts
```diff
@@ -0,0 +1,126 @@
...
+  public buttonDisabled(): boolean {
+    return this._buttonElement?.disabled ?? true;
+  }
+
+  public setDisabled(value: boolean): void {
+    toggleClass(this._rootElement, value, BUTTON_AREA_CONSTANTS.classes.DISABLED);
+    
+    if (this._buttonElement) {
+      this._buttonElement?.toggleAttribute(BUTTON_AREA_CONSTANTS.attributes.DISABLED, value);
+    }
```
DRiFTy17, line 55: Minor but you don't need the wrapper `if` statement here.

DRiFTy17 (commented):
This is looking great so far! A few very minor comments below.

One other thing I noticed was that when clicking on the icon-button within the button-area, it also triggered the ripple animation on the button area as well. We will probably want to avoid that?

Also, for demo purposes when it comes time for docs maybe we could use this in an expansion panel with the accessibility attributes hooked up to come full circle on that pattern?

src/lib/index.ts
```diff
@@ -139,6 +141,7 @@ export * from './utils';
...
```
DRiFTy17, line 144: Duplicate export statement

src/dev/pages/button-area/button-area.ejs
```diff
@@ -0,0 +1,46 @@
...
+      <button slot="button" id="expansion-panel-button" aria-labelledby="expansion-panel-heading" aria-controls="expansion-panel-content" aria-expanded="false"></button>
+      <div class="flex-wrapper">
+        <div>
+          <div class="forge-typography--headline5" id="expansion-panel-heading">Expansion panel</div>
+          <div>Subheading</div>
+        </div>
+        <forge-icon-button data-forge-ignore>
+          <button>
+            <forge-icon name="favorite"></forge-icon>
+          </button>
```
DRiFTy17, line 34: I noticed in this demo when tabbing to this inner button and pressing the space key that it toggled the expansion panel, but clicking the button did not. Is that intentional?

samrichardsontylertech: Yeah, it's demonstrating the ignore attribute

DRiFTy17 (approved):

PR #365, opened 2023-08-29
feat(popup): add popup ref to close event, emit from host

MikeMatusz:
## PR Checklist

## Describe the new behavior?
The `forge-popup-close` event now contains a reference to the `forge-popup` element in the event `detail`, so that it's possible to distinguish which popup is responsible for the event in the case where there are multiple (e.g. when clicking a menu item shows a popup).  Also emitting the close event from the host element, in addition to the target element, which also ensures a one-to-one relationship between the event and popup given you have a reference to it.  The latter might be more useful in JavaScript, while the former method disambiguates the events in the declarative use case.

## Additional information
Would it be better to create a new event to raise instead of re-using the same event as the target?  If so, what should it be called?  `forge-popup-host-closed`?


DRiFTy17 (approved):
This looks good to me! I think emitting the same event name from both elements is fine personally. It's kind of non-standard in regards to typical Forge patterns for uniqueness, but this is a non-standard scenario anyway, not to mention it's common practice in HTML to reuse event names across elements. I'm ok with it if everyone else is.

PR #367, opened 2023-08-29
feat(keyboard-shortcut): add activate callback

samrichardsontylertech:
## PR Checklist

## Describe the new behavior?
This adds a new `activateCallback` property as an alternative to requiring binding to the `keyboard-shortcut-activate` event to use the component. There are some situations where this method could be more concise than adding an event listener.

### DIFF
--- a/src/lib/keyboard-shortcut/keyboard-shortcut-foundation.ts
+++ b/src/lib/keyboard-shortcut/keyboard-shortcut-foundation.ts
@@ -84,7 +84,7 @@ export class KeyboardShortcutFoundation implements IKeyboardShortcutFoundation {
         evt.preventDefault();
       }
       this._adapter.emitHostEvent(KEYBOARD_SHORTCUT_CONSTANTS.events.ACTIVATE, evt);
-      this._activateCallback?.(evt);
+      this._activateCallback?.call(null, evt);
     }
   }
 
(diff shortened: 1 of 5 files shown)


src/lib/keyboard-shortcut/keyboard-shortcut-foundation.ts
```diff
@@ -194,4 +197,14 @@ export class KeyboardShortcutFoundation implements IKeyboardShortcutFoundation {
...
   }
+
+  /** Gets/sets the activation callback. */
+  public get activateCallback(): KeyboardShortcutActivateCallback | null | undefined {
+    return this._activateCallback;
+  }
+  public set activateCallback(value: KeyboardShortcutActivateCallback | null | undefined) {
+    if (this._activateCallback !== value) {
+      this._activateCallback = value;
+    }
```
DRiFTy17, line 208: Just a side note, but since this doesn't have the possibility of creating an infinite loop like other properties that set host attributes, I don't typically wrap these in `if` statements.

src/lib/keyboard-shortcut/keyboard-shortcut-foundation.ts
```diff
@@ -82,6 +84,7 @@ export class KeyboardShortcutFoundation implements IKeyboardShortcutFoundation {
         evt.preventDefault();
       }
       this._adapter.emitHostEvent(KEYBOARD_SHORTCUT_CONSTANTS.events.ACTIVATE, evt);
+      this._activateCallback?.(evt);
```
DRiFTy17, line 87: It may be worth invoking the callback with `.call(null, evt)` or `.call({}, evt)` to ensure the internal context isn't provided to the consumer.

DRiFTy17 (approved):

PR #402, opened 2023-10-06
fix(paginator): fixed how `offset` updates `pageIndex` along with updates to focus management

### LINKED ISSUE #400: [paginator] initializing `offset` before `total` does not adjust `pageSize`
**Describe the bug:**
When initializing values on the paginator for the first time (when total starts at `0` to `>0` value), if setting the `offset` property prior to setting the `total` property, the `pageSize` will not be set. This is more obvious when using frameworks like Angular where the bindings could be set in any order and developers shouldn't have to worry about the order of their bindings...

**To Reproduce:**
1. Create a paginator using `document.createElement('forge-paginator')`
2. Set the `offset` property on the new paginator instance to a value that matches the page size
3. Set the `total` property to `100`
4. Append the paginator instance to the DOM
5. See that the page size hasn't changed

**Expected behavior:**
The `offset` value should be allowed to be set regardless of when `total` is set.

 - Forge version: `^2.0.0`
 - Browser: All
 - Platform: All
 - OS: All

**Additional context:**
We may just need to check if `offset` is set to anything when `total` is set and recalculate the `pageSize` in the `total` setter as well to make sure that order doesn't matter.

### LINKED ISSUE #403: [paginator] focus being set to paginator buttons when state changes
**Describe the bug:**
This was caused by #398.

The changes in #398 cause the buttons to receive focus when the state changes programmatically, which is incorrect. The focus needs to only move in response to a user interaction with the paginator.

DRiFTy17:
## PR Checklist

## Describe the new behavior?
The paginator will now allow for the `offset` value to be set **before** the `total` property. Essentially anytime the `total` property go from `0` to `>0` and there is an `offset` value applied, the `pageIndex` will now properly update.

Additionally, changes were made to how component state updates the DOM when various properties are changes to allow for fixing a bug related to how focus is managed and moved around within the component for actions triggered by user interaction vs programmatic state updates.

## Additional information
Fixes #400 
Fixes #403

### DIFF
--- a/src/lib/paginator/paginator-foundation.ts
+++ b/src/lib/paginator/paginator-foundation.ts
@@ -1,8 +1,7 @@
 import { coerceNumber, ICustomElementFoundation, isArray, isDefined } from '@tylertech/forge-core';
-import { IPaginatorAdapter } from './paginator-adapter';
+import { IPaginatorAdapter, PaginatorFieldIdentifier } from './paginator-adapter';
 import { PaginatorAlternativeAlignment, PAGINATOR_CONSTANTS, IPaginatorChangeEvent } from './paginator-constants';
-import { ISelectOption, ISelectComponent } from '../select';
-
+import { ISelectOption } from '../select';
 
 export interface IPaginatorFoundation extends ICustomElementFoundation {
   pageIndex: number;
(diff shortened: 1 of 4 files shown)


src/lib/paginator/paginator-foundation.ts
```diff
@@ -82,24 +83,24 @@ export class PaginatorFoundation {
...
+    return this._offset;
   }
 
   /** The total number of items to be paginated. Default is 0. */
   public set total(value: number) {
     if (this._total !== value) {
       this._total = value;
+
+      if (this._offset > 0 && this._total > 0) {
+        this._computePageIndexFromOffset(this._offset);
```
MikeMatusz, line 101: Do we need to recalculate the page index if the total goes back to zero (e.g. if used for search results, the user does a new search with no results).  Or would we also expect offset to get set at the same time?

DRiFTy17: Good question. I believe those are typically expected to be managed by the consumer at the same time since they will likely have bindings to both `total` and `pageIndex`/`offset` at the same time.

In the past we haven't been updating the `pageIndex` in this scenario because it could potentially cause the values to become out of sync with the data model in the consumer if we update on their behalf.

Does that sound like it matches your expectations?

DRiFTy17:
Just an FYI, but this PR was updated to also include some other fixes related to `offset`, and this included a larger refactor that was closely related to the focus issue so I decided to combine the two into this single PR.

src/lib/paginator/paginator-foundation.ts
MikeMatusz: Yep, makes sense to me now.

MikeMatusz (approved):

PR #409, opened 2023-10-14
feat(checkbox): refactor checkbox component

### LINKED ISSUE #370: [@next] refactor checkbox component
Refactor the checkbox component to use new token and style patterns, while removing dependency on MDC, and integrate with (ElementInternals) APIs.

Migrate tests to WTR, update docs, and usages across library.

samrichardsontylertech:
## PR Checklist

## Describe the new behavior?
Removed Material dependencies from the checkbox component, added a toggle method, added tokens and new animations, set up element internals, refined the base form interface/class and forwarding/reflection utilities, and fixed related bugs around the state layer, focus indicator, and switch components.

Closes #370


src/lib/switch/switch.scss
```diff
@@ -12,19 +12,7 @@
   display: none;
 }
 
-:host([disabled]) {
-  @include core.host-disabled;
-
-  .forge-switch {
-    pointer-events: none;
```
DRiFTy17, line 19: We may have lost this style during this refactor (related to my comment about disabled on the demo)

src/lib/checkbox/checkbox.ts
```diff
@@ -18,66 +25,269 @@ declare global {
...
+ * @attribute {string} value - The value of the checkbox when checked.
+ * @attribute {string} disabled - Controls if the switch is disabled.
+ * @attribute {string} required - Controls if the switch is required.
+ * @attribute {string} readonly - Controls if the switch is readonly.
+ * @attribute {string} dense - Sets the density state.
+ * @attribute {string} label-position - Sets whether the label appears before or after the switch.
+ * 
+ * @method {(force?: boolean) => void} toggle - Toggles whether the checkbox is is checked or forces a checked state.
+ *  
+ * @event change {CustomEvent} - Dispatches when the switch's value changes.
```
DRiFTy17, line 60: The type should come before the name here, and technically it's not a `CustomEvent` since we're redispatching, right?

src/lib/checkbox/checkbox-foundation.ts
```diff
@@ -1,209 +1,194 @@
...
+    return this._checked ? this._value : null;
+  }
+  private get _formState(): CheckboxState {
+    return this._checked
+      ? this._indeterminate
+        ? 'checked-indeterminate'
+        : 'checked'
+      : this._indeterminate
+        ? 'unchecked-indeterminate'
+        : 'unchecked';
```
DRiFTy17, line 42: Should we move these strings to constants?

src/lib/checkbox/checkbox.test.ts
```diff
@@ -0,0 +1,451 @@
...
+import { expect } from '@esm-bundle/chai';
+import { fixture, html } from '@open-wc/testing';
+import { getShadowElement } from '@tylertech/forge-core';
+import { sendKeys, sendMouse } from '@web/test-runner-commands';
+import { spy } from 'sinon';
+import { TestHarness } from '../../test/utils/test-harness';
+import { CHECKBOX_CONSTANTS, ICheckboxComponent } from '../checkbox';
+import { IFocusIndicatorComponent } from '../focus-indicator';
+import { IStateLayerComponent } from '../state-layer';
+import exp from 'constants';
```
DRiFTy17, line 12: Looks like we can remove this import since it was likely added by mistake (I do this all the time)

src/lib/checkbox/checkbox.test.ts
```diff
@@ -0,0 +1,451 @@
+/// <reference types="mocha" />
```
DRiFTy17, line 1: It would be worth removing this I think. It won't be needed in the future, and it will help with not accidentally leaving an `it.only` in here.

src/lib/checkbox/checkbox.scss
```diff
@@ -1,14 +1,149 @@
...
+  .forge-checkbox {
+    .container {
+      @include core.container-dense;
+    }
+  }
+}
+
+:host([disabled]) {
+  .forge-checkbox {
+    @include core.checkbox-disabled;
```
DRiFTy17, line 112: Should the label also show the `not-allowed` cursor when disabled? Same was noticed in the switch as well.

src/lib/checkbox/checkbox-foundation.ts
```diff
@@ -1,209 +1,194 @@
...
+      this._adapter.setChecked(this._checked);
+      evt.stopPropagation();
       return;
     }
 
-    const handler: TimerHandler = () => {
-      this._adapter.removeRootClass(this._currentAnimationClass);
-      this._enableAnimationEndHandler = false;
-    };
+    this._adapter.redispatchEvent(evt);
```
DRiFTy17, line 76: The `event.target.checked` property on this event doesn't appear to reflect the "new" state in the listener. We should be able to query the target element from the listener for its checked state and receive the new state.

Also, it appears that we cannot cancel the `input` or `change` event (since the native events do not allow that). It seems that we should be consistent with the switch here and allow for `preventDefault()`?

src/lib/core/base/base-adapter.ts
```diff
@@ -38,6 +39,20 @@ export class BaseAdapter<T extends IBaseComponent> implements IBaseAdapter {
     toggleAttribute(this._component, hasAttribute, name, value);
   }
 
+  public redispatchEvent(event: Event): boolean {
+    const isFromLightDom = !((event.target as HTMLElement)?.getRootNode() instanceof ShadowRoot);
```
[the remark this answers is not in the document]
samrichardsontylertech: The name could be clearer but this shouldn't cause any errors because the next step checks whether the event is composed. In practice this should prevent cross-shadow root usage.

src/lib/switch/switch.scss
samrichardsontylertech: This was purposeful, to make sure that the disabled cursor is set. I fixed the disabled interactions by only applying active styles to non-disabled or readonly hosts.

src/lib/checkbox/_core.scss
```diff
@@ -0,0 +1,159 @@
...
+  // Animate from right to left in RTL
+  &:dir(rtl) {
+    stroke-dashoffset: 40;
+  }
+}
+
+@mixin label {
+  @include typography.style(label);
+
+  margin-inline: 8px;
```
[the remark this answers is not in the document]
samrichardsontylertech: I don't even know why this is here. It should be handled by the gap token at the root element.

src/lib/checkbox/checkbox-foundation.ts
samrichardsontylertech: I thought they made more sense as string literals 🤷‍♂️

src/lib/checkbox/checkbox.scss
samrichardsontylertech: I personally think it shouldn't for consistency with external labels and the loose visual connection the label has to the input, but I could be convinced otherwise

src/lib/checkbox/checkbox-foundation.ts
samrichardsontylertech: I updated `redispatchEvent` to accept option overrides for situations like this

src/lib/core/styles/tokens/checkbox/_tokens.scss
```diff
@@ -0,0 +1,65 @@
...
+  icon-color: utils.module-val(checkbox, icon-color, theme.variable(on-tertiary)),
+  state-layer-size: utils.module-val(checkbox, state-layer-size, 40px),
+  state-layer-dense-size: utils.module-val(checkbox, state-layer-size, 24px),
+
+  // Checkbox
+  background-color: utils.module-val(checkbox, background-color, theme.variable(surface)),
+  width: utils.module-ref(checkbox, width, size),
+  height: utils.module-ref(checkbox, height, size),
+  unchecked-border-width: utils.module-ref(checkbox, unchecked-border-width, border-width),
+  unchecked-border-color: utils.module-val(checkbox, unchecked-border-color, theme.variable(surface-container-high)), // TODO: check out surface-container
```
DRiFTy17, line 24: Leftover comment

src/lib/checkbox/checkbox.scss
```diff
@@ -1,14 +1,149 @@
-@use '../theme';
-@use './checkbox.mixins' as mixins;
-@use '../form-field/mixins' as form-field-mixins;
-
-@include mixins.core-styles;
-@include form-field-mixins.shadow-styles;
+@use './core';
+@use './configuration';
+@use '../core/styles/tokens/checkbox/tokens';
```
DRiFTy17, line 3: Let's remove this tokens import so it doesn't give the impression we should be using tokens directly here

src/lib/checkbox/checkbox.scss
DRiFTy17, line 1: Should we implement `forced-colors: active` here?

src/lib/checkbox/checkbox.html
```diff
@@ -1,15 +1,22 @@
...
+      </slot>
+      <div class="background" part="background">
+        <svg class="icon icon--checked" aria-hidden="true" viewBox="0 0 24 24" part="checkmark">
+          <path class="path path--checkmark" d="M1.73,12.91 8.1,19.28 22.79,4.59" part="checkmark-path"></path>
         </svg>
-        <div class="mdc-checkbox__mixedmark" part="mixedmark"></div>
+        <svg class="icon icon--indeterminate" aria-hidden="true" viewBox="0 0 24 24" part="mixedmark">
+          <line x1="2" y1="12" x2="22" y2="12" part="mixedmark-path"></line>
+        </svg>
+        <forge-focus-indicator target="container" part="focus-indicator"></forge-focus-indicator>
```
DRiFTy17, line 14: For the `<forge-focus-indicator>` and `<forge-state-layer>` elements we should use `exportparts`. See the slider and tab for an example.

src/lib/checkbox/checkbox.html
```diff
@@ -1,15 +1,22 @@
...
-        <svg class="mdc-checkbox__checkmark" viewBox="0 0 24 24" part="checkmark-svg">
-          <path class="mdc-checkbox__checkmark-path" fill="none" d="M1.73,12.91 8.1,19.28 22.79,4.59" part="checkmark-path"></path>
+  <div class="forge-checkbox" part="checkbox">
+    <div id="container" class="container" part="input-container">
+      <slot name="input">
+        <input id="input" class="input" type="checkbox" part="input" />
+      </slot>
+      <div class="background" part="background">
+        <svg class="icon icon--checked" aria-hidden="true" viewBox="0 0 24 24" part="checkmark">
+          <path class="path path--checkmark" d="M1.73,12.91 8.1,19.28 22.79,4.59" part="checkmark-path"></path>
```
DRiFTy17, line 9: I'm torn on whether the `<path>` and `<line>` elements should be exposed as `part` elements in this template, thoughts?

My only hesitation is that if we expose everything then it's part of the public API of the element. Are all of these elements we expect developers to need access to or can they be hidden implementation details?

DRiFTy17 (commented):
Things are looking good, just a couple questions/comments:

When a slotted `<input>` is used, can we have it automatically add the `slot="input"` attribute when initializing the adapter if one doesn't exist? This makes it easier to use, but on top of that we can improve performance because it can immediately initialize against the slotted `<input>` instead as well regardless:
   ```ts
   const slottedInput = this._component.querySelector('input[type=checkbox]:not([slot])') as HTMLInputElement;
    if (slottedInput) {
      slottedInput.slot = 'input';
    }
   ```

Additionally, I know you're probably waiting until the end, but we should update the delegate as well included tests for anything that is missing to improve coverage.

src/lib/checkbox/checkbox.html
samrichardsontylertech: I kept the `part` attributes here just because the old checkbox had them, but I don't think it makes a lot of sense. They aren't even targeted by an of the internal styles.

DRiFTy17 (approved):

PR #416, opened 2023-10-23
fix(table): sortable column headers will now use `<button>` element for a11y purposes

### LINKED ISSUE #63: [table] Accessibility: Table headers cannot be accessed/triggered from keyboard
**Describe the feature:**
When using the keyboard to access the table, header and its controls are not reachable. This makes screenreaders and keyboard navigation skip to the content of the table. It would be nice if we can access this controls from keyboard input.
Also aria labels

**Additional context:**
The table currently handles mouse interactions (it uses mousedown and mouseup events to trigger this).
Screen Shot 2022-05-31 at 8 35 36 PM

DRiFTy17:
## PR Checklist

## Describe the new behavior?
Sortable table headers will now use a `<button>` element to ensure they are part of the tab order to fix accessibility related issues to triggering sort changes on the columns via keyboard only.

Following W3 table guidance for sortable columns here:
https://www.w3.org/WAI/ARIA/apg/patterns/table/examples/sortable-table/

We should plan to introduce an API for setting the table `<caption>` as well for improved screen reader support.

## Additional information
Fixes #63

### DIFF
--- a/src/lib/table/table-foundation.ts
+++ b/src/lib/table/table-foundation.ts
@@ -105,6 +105,7 @@ export class TableFoundation implements ITableFoundation {
   private _rowDoubleClickListener: (evt: Event) => void;
   private _selectRowListener: (evt: Event) => void;
   private _selectAllListener: (evt: Event) => void;
+  private _sortableHeadCellKeydownListener: EventListener;
   private _headRowMouseDownListener: (evt: MouseEvent) => void;
   private _headRowContextMenuListener: (evt: MouseEvent) => void;
   private _documentMouseMoveListener: (evt: MouseEvent) => void;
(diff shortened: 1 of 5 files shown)


src/lib/table/table-foundation.ts
```diff
@@ -979,6 +986,16 @@ export class TableFoundation implements ITableFoundation {
     this._headCellMouseDownIndex = undefined;
   }
 
+  private _onSortableHeadCellKeydown(evt: KeyboardEvent): void {
```
samrichardsontylertech, line 989: Should this use click events instead of keydown and mousedown now that it's a button?

samrichardsontylertech (approved):

PR #422, opened 2023-11-08
chore(list-item): improve anchor handling and semantics

DRiFTy17:


samrichardsontylertech (approved):

DRiFTy17:
To avoid issues with semantics I'm going to close this PR in favor of #423 instead, which has similar changes but doesn't refactor everything to the host. We may perform further iteration in the future as a hybrid of both of these approaches.

PR #423, opened 2023-11-08
chore(list-item): improve anchor handling and semantics

DRiFTy17:


samrichardsontylertech (approved):

DRiFTy17:
More work may need to be done here. See #422 for an alternate approach to move the semantics to the host element. We chose not to go that far for now, but will revisit accessibility-related concerns in the list in general soon.

PR #431, opened 2023-11-15
feat(icon-button): refactored icon-button

DRiFTy17:


samrichardsontylertech (commented):
Looks really good overall. The only piece that looks weird to me is the inline styles.

samrichardsontylertech (approved):

PR #434, opened 2023-11-20
feat(radio): refactor radio component

### LINKED ISSUE #371: [@next] refactor radio component
Refactor the radio component to remove reference to MDC styles. Update to use new token and style patterns, and integrate with (ElementInternals) APIs.

While doing so, we should look to create a new `<forge-radio-group>` parent element to make usage of this component less confusing.

Update docs, tests in WTR, and usages across library.

samrichardsontylertech:
## PR Checklist

## Describe the new behavior?
Refactor of the radio component to remove Material dependencies, improve accessibility, and adhere more closely to native input behavior.

Closes #371


src/lib/core/utils/mixin-utils.ts
DRiFTy17, line 1: Something along the lines of this should get us started:
```typescript
import { MixinBase } from '../../constants';

type MixinFunction<T extends MixinBase = MixinBase, R extends T = T & MixinBase> = (Base: T) => R;
type UnionToIntersection<T> = (T extends any ? (x: T) => any : never) extends (x: infer R) => any ? R : never;
type MixinReturnValue<T extends MixinFunction<any, any>[]> = UnionToIntersection<{ [K in keyof T]: T[K] extends MixinFunction<any, infer U> ? U : never }[number]>;

export function combineMixins<T extends MixinBase, K extends MixinFunction<T, any>[]>(base: T, ...mixins: K) {
  return mixins.reduce((acc, cur) => cur(acc), base) as MixinReturnValue<K>;
}
```

src/lib/radio/radio/radio.html
```diff
@@ -0,0 +1,12 @@
+<template>
+  <div id="radio" class="forge-radio" part="radio">
+    <div id="container" class="container" aria-hidden="true">
+      <div id="background" class="background" part="background"></div>
+      <forge-focus-indicator target=":host" exportparts="indicator:focus-indicator"></forge-focus-indicator>
```
DRiFTy17, line 5: Just an FYI for when you merge in the latest changes from the app-bar PR (assuming it is merged before this), the template for `<forge-focus-indicator>` was removed so we can also remove `exportparts` and replace it with just `part="focus-indicator"`

src/lib/radio/radio/radio-foundation.ts
```diff
@@ -0,0 +1,233 @@
...
+        this._adapter.focusPrevious();
+        break;
+    }
+  }
+
+  private _handleKeyup(evt: KeyboardEvent): void {
+    if (evt.key === ' ') {
+      evt.preventDefault();
+      this._activate(evt);
+    }
```
DRiFTy17, line 104: When testing the component I noticed that when no radios were checked, tabbing into the first one, and pressing space did not seem to check the radio.

src/lib/radio/radio/radio-constants.ts
```diff
@@ -0,0 +1,53 @@
...
+
+export const RADIO_CONSTANTS = {
+  elementName,
+  attributes,
+  selectors,
+  events,
+  observedAriaAttributes
+};
+
+export const tryCheck = Symbol('tryCheck');
```
DRiFTy17, line 50: Maybe comment what this `Symbol` is used for either in the class implementation or here? Had to look it up when I first saw it just to be sure I was understanding.

src/lib/radio/radio/index.scss
DRiFTy17, line 1: Just a reminder to not forget the forwards here

src/lib/core/base/base-component.ts
```diff
@@ -1,3 +1,8 @@
 export interface IBaseComponent extends HTMLElement {}
 
-export abstract class BaseComponent extends HTMLElement implements IBaseComponent {}
+export abstract class BaseComponent extends HTMLElement implements IBaseComponent {
+  // We provide default implementation of these methods so that they can be extended by other
+  // classes.
+  public connectedCallback(): void {}
+  public attributeChangedCallback(name: string, oldValue: string | null, newValue: string | null): void {}
```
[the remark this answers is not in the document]
samrichardsontylertech: These are only in the class so we can call `super()` from the mixins without worrying about which one ends up being the parent class. I don't think adding them to the interface would solve that problem?

src/lib/radio/radio/radio-constants.ts
samrichardsontylertech: Will do. I'm also concerned that the name is generic enough that we could run into conflicts in the future. I might change it to something more unique to the radio or move it into the top level constants to allow other components to use it.

src/lib/core/base/base-component.ts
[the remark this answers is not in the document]
DRiFTy17: Oh yup that makes sense to me. I'm fine with this.

PR #442, opened 2023-12-12
[@next] refactor button toggle

DRiFTy17:


samrichardsontylertech (changes requested):
Looks good except for color contrast with some of the themes. Secondary fails WCAG AA and several of the others are borderline, only passing AA large text criteria. I don't think the text is bold enough to qualify as large text here.

samrichardsontylertech (approved):
We can tweak the theme colors in another PR, otherwise this looks good

PR #451, opened 2024-01-16
[@next] refactor avatar to use design tokens

MikeMatusz:
## PR Checklist

- Does this PR introduce a breaking change? Y (autoColor property is removed, css variables changed)

## Describe the new behavior?
Converted `forge-avatar` to use design tokens.  Removed `autoColor` property due to accessibility concerns, and switched to using a class to hide the background color when an image is loading/loaded rather than storing the previous background color, which could result in getting stuck as "inherit" if render is called again while an image is still loading.  

## Additional information
There's still a race condition around image loading, but I'll create a separate issue to track that, because it's something of an edge case, and was resulting in a bit of complexity to try and address.  Not sure the time or extra code is worth it at this point.  The main remaining symptom would be if you set imageUrl to an image that loads slowly, and then set it to a different imageUrl that loads before the first one finishes, the first imageUrl will end up being applied over the second.


src/lib/core/styles/tokens/avatar/_tokens.scss
```diff
@@ -0,0 +1,24 @@
+@use 'sass:map';
+@use '../../animation';
+@use '../../border';
+@use '../../elevation';
+@use '../../shape';
+@use '../../spacing';
+@use '../../theme';
+@use '../../typography';
```
DRiFTy17, line 8: Let's remove any unused modules from here. Not critical, but trying to keep these clean so we can quickly see what global tokens are inherited.

src/test/spec/avatar/avatar.spec.ts
DRiFTy17, line 1: Just a side note about the conversion of tests from Karma to WTR: I've done both ways of refactoring them by copying directly to WTR and updating the syntax for mocha compatibility, as well as just plain rewriting the tests because the older Karma tests were not really that great...

These tests seem to be decent at first glance, but just wanted to mention that it's worth taking the time to audit these and make sure they are clean, concise, and doing what they should be. Also, make sure to take a look at the code coverage for anything we might have missed.

src/lib/avatar/avatar.ts
```diff
@@ -25,6 +24,27 @@ declare global {
  * The custom element class behind the `<forge-avatar>` element.
```
DRiFTy17, line 24: I've been removing these useless comments and using `@summary` and/or `@description` with something useful.

src/lib/core/styles/tokens/avatar/_tokens.scss
```diff
@@ -0,0 +1,24 @@
...
+@use '../../utils';
+
+$tokens: (
+  size: utils.module-val(avatar, size, 40px),
+  font-size: utils.module-val(avatar, font-size, 1rem),
+  font-weight: utils.module-val(avatar, font-weight, 400),
```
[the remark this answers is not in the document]
MikeMatusz: I think I agree with letting the standard typography tokens handle it, because those always specify size/weight/letter-spacing/line-height together, and if we just allow setting size/weight via variables here, the other values might be off.  If you have to override the font properties, it makes sense to consider them holistically, which I feel like is better encouraged by not providing these at the component level.

src/lib/avatar/_core.scss
```diff
@@ -0,0 +1,32 @@
...
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  box-sizing: border-box;
+  background-position: center;
+  background-repeat: no-repeat;
+  background-size: cover;
+  overflow: hidden;
+  transition: height #{token(transition-duration)} #{token(transition-timing)}, 
+              width #{token(transition-duration)} #{token(transition-timing)};
```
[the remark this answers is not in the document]
MikeMatusz: Going to take my best stab at it, but I'm seeing a few minor inconsistencies between those two references, and even compared to a few existing _core.scss files with grouping and similar properties, so definitely after I push again, please let me know if you'd have grouped/ordered something differently.  I definitely like the principles of idiomatic and concentric CSS and being consistent with the ordering in a way that makes CSS easier to understand at a glance, but definitely looking forward to when we can apply it via a formatter.

samrichardsontylertech (commented):
Kieran covered all my thoughts pretty well. This looks good, it just needs some cleanup.

DRiFTy17 (changes requested):
Everything looks really good, but I think we should switch to the `subheading2` typography for consistency with the 2.x design.

DRiFTy17 (approved):

PR #453, opened 2024-01-17
[@next] tooltip refactor

DRiFTy17:
- Refactor the tooltip to use `<forge-overlay>`
- Provides static/inline HTML support for custom content
- Improved a11y support with the new `type` attribute
- Backwards-compatible with 2.x usage and API
  - No longer sets `aria-describedby` by default. The element is considered presentational by default now
- Refactored tests to WTR
- Implement design tokens


src/dev/pages/button-area/button-area.ejs
```diff
@@ -29,7 +29,7 @@
         <forge-icon-button data-forge-ignore>
           <forge-icon name="favorite"></forge-icon>
         </forge-icon-button>
-        <forge-tooltip>Like</forge-tooltip>
+        <forge-toolti type="label">Like</forge-tooltip>
```
samrichardsontylertech, line 32: You've got a typo here

samrichardsontylertech (approved):

MikeMatusz (approved):

PR #454, opened 2024-01-19
[@next] port deprecated (2.x) button for backwards compatibility

DRiFTy17:
This change brings support for the "decorator" pattern for buttons that we used in the 2.x version to the 3.0 version. This is only for backwards compatibility and _should_ be API-compatible with the 2.x implementation.

It allows for slotting native `<button>` elements in for backwards compatibility to ease the upgrade process:
```html
<forge-deprecated-button>
  <button type="button">Decorated button</button>
</forge-deprecated-button>
```

This was ported from the 2.x version and integrated with design tokens (it uses the same styles from the new 3.0 button but with `::slotted()` selectors instead and some minor overrides to match 2.x.

This is not a direct port from 2.x and it wasn't intended to be. The styles were moved to the shadow DOM, and it was integrated with the new state-layer and focus-indicators. This is purely for supporting slotted native `<button>` elements.

Consumers will need to opt-in to renaming their existing elements to `<forge-deprecated-button>` if they want to avoid sweeping their code.

### DIFF
--- a/src/dev/pages/button/button.ejs
+++ b/src/dev/pages/button/button.ejs
@@ -109,6 +109,40 @@
       </form>
     </dialog>
   </section>
+
+  <section class="vert">
+    <h3 class="forge-typography--heading2">Deprecated (legacy) button</h3>
+    <forge-deprecated-button>
+      <button type="button">Default button</button>
+    </forge-deprecated-button>
+    <forge-deprecated-button type="outlined">
+      <button type="button">Outlined button</button>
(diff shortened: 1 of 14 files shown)


samrichardsontylertech (approved):

MikeMatusz (approved):
Looks good.  Probably should throw an `@deprecated` JSDoc tag on the component class/interface and define method.  Curious if there'd be a good way to propagate that to the Angular modules via the schematic, too.

PR #456, opened 2024-01-19
[@next] create low level field component

### LINKED ISSUE #455: [@next] create low-level/reusable "form field" component
The idea behind this is to encapsulate the look-and-feel of the text-field and select into a reusable component that can be used in both elements.

This component should:
- [ ] Remove dependency on MDC
- [ ] Introduce design tokens
- [ ] Convert to shadow DOM
- [ ] Accept both an `<input>` and `<textarea>` as well as just a label for non-interactive entry
- [ ] Write tests with WTR
- [ ] Update docs

samrichardsontylertech:
## PR Checklist

## Describe the new behavior?
This adds a new field component to replace the existing field component and form field styles. It provides a reusable template for fields (text-field, select, etc.) that share common layouts, styles, and basic behavior.

## Additional information
Closes #455


src/lib/core/styles/tokens/typography/_tokens.label.scss
```diff
@@ -10,6 +10,13 @@ $label: utils.inherit-map(core.$base, (
   letter-spacing: type-utils.calc-letter-spacing(0.5, scale.value('0875'))
 )) !default;
 
+// TODO: solidify label 2 styles
+$label2: utils.inherit-map(core.$base, (
+  font-size: type-utils.font-size-relative(label, font-size, '0813'),
+  line-height: type-utils.font-size-relative(label, line-height, '1250'),
```
DRiFTy17, line 16: We may want to use a `normal` line height on this to ensure wrapping labels aren't too space apart. I'd also consider doing this on the standard "label" style as well?

src/lib/field-next/_core.layout.scss
```diff
@@ -0,0 +1,297 @@
...
+
+@mixin slotted-floating-input {
+  padding-block-start: calc($floating-offset * 2) !important; // TODO: find a way to remove this !important
+}
+
+@mixin floating-animation($animation-name) {
+  animation-name: #{$animation-name};
+  animation-duration: #{token(inset-animation-duration)};
+  animation-timing-function: #{token(inset-animation-timing-function)};
+}
```
DRiFTy17, line 265: I'd be curious what it looks like to animate the label when unfloating. It may appear a bit more smooth after clearing text. If this isn't trivial to add we can add a task for looking into this later. I only suggest this because our old floating label implementation does animate back to its unfloated/resting position when blurring the field.

src/lib/field-next/field.ts
```diff
@@ -0,0 +1,203 @@
...
+      FIELD_CONSTANTS.attributes.VARIANT,
+      FIELD_CONSTANTS.attributes.THEME,
+      FIELD_CONSTANTS.attributes.SHAPE,
+      FIELD_CONSTANTS.attributes.DENSITY,
+      FIELD_CONSTANTS.attributes.DENSE,
+      FIELD_CONSTANTS.attributes.POPOVER_ICON,
+      FIELD_CONSTANTS.attributes.POPOVER_EXPANDED,
+      FIELD_CONSTANTS.attributes.MULTILINE,
+      FIELD_CONSTANTS.attributes.SUPPORT_TEXT_INSET
+    ];
```
DRiFTy17, line 88: Let's use our new `observedAttributes` pattern here so new observed attributes can be added automatically.

src/lib/field-next/_core.scss
```diff
@@ -0,0 +1,189 @@
...
+  @include override('main-content-padding-start', #{token(container-padding-start)}, value);
+  @include override('main-content-padding-end', #{token(container-padding-end)}, value);
+
+  display: grid;
+  position: relative;
+}
+
+@mixin field-disabled {
+  opacity: #{token(disabled-opacity)};
+}
```
DRiFTy17, line 34: Should we add the `not-allowed` cursor style when disabled?

src/lib/field-next/_core.shape.scss
```diff
@@ -0,0 +1,20 @@
...
+  }
+}
+
+@mixin shape($shape) {
+  @if $shape == 'squared' {
+    @include override('shape', 0, value);
+  } @else if $shape == 'rounded' {
+    @include override('shape', shape.variable(full), value);
+    @include override('container-padding-start', 'pill-container-padding-start');
+    @include override('container-padding-end', 'pill-container-padding-end');
```
DRiFTy17, line 18: We may want to introduce some inline padding in the rounded variant to align the content with the half circle.

src/lib/core/styles/tokens/field/_tokens.scss
```diff
@@ -0,0 +1,135 @@
+@use 'sass:map';
+@use '../../animation';
+@use '../../border';
+@use '../../elevation';
+@use '../../shape';
+@use '../../theme';
+@use '../../utils';
+
+$tokens: (
+  display: utils.module-val(field, display, block),
```
DRiFTy17, line 10: I think we can get rid of this `display` token as it's likely more of an internal layout style.

src/lib/core/styles/tokens/field/_tokens.scss
```diff
@@ -0,0 +1,135 @@
...
+  pill-container-padding: utils.module-val(field, pill-container-padding, 16px),
+  pill-container-padding-start: utils.module-ref(field, pill-container-padding-start, pill-container-padding),
+  pill-container-padding-end: utils.module-ref(field, pill-container-padding-end, pill-container-padding),
+  
+  // Sizes
+  height-extra-small: utils.module-val(field, height-extra-small, 24px),
+  height-small: utils.module-val(field, height-small, 32px),
+  height-medium: utils.module-val(field, height-medium, 40px),
+  height-large: utils.module-val(field, height-large, 48px),
+  height-extra-large: utils.module-val(field, height-extra-large, 56px),
```
DRiFTy17, line 49: Let's create some base-level tokens like `height` in this case to allow for adjusting a single token across all variants if desired, while still supporting individual variant-specific token overrides.

See other comment about needing to update our `override()` function implementation to support this, but our code should be structured this way to make it easier on consumers for overriding.

src/lib/field-next/field-foundation.ts
```diff
@@ -0,0 +1,268 @@
...
+  private _variant: FieldVariant = FIELD_CONSTANTS.defaults.DEFAULT_VARIANT;
+  private _theme: FieldTheme = FIELD_CONSTANTS.defaults.DEFAULT_THEME;
+  private _shape: FieldShape = FIELD_CONSTANTS.defaults.DEFAULT_SHAPE;
+  private _density: FieldDensity = FIELD_CONSTANTS.defaults.DEFAULT_DENSITY;
+  private _dense = false;
+  private _popoverIcon = false;
+  private _popoverExpanded = false;
+  private _multiline = false;
+  private _supportTextInset: FieldSupportTextInset = FIELD_CONSTANTS.defaults.DEFAULT_SUPPORT_TEXT_INSET;
+  private _isInitialized = false;
```
DRiFTy17, line 42: You should try replacing this property with `this._adapter.isConnected` instead unless it relies on timing of internal setup instead of just being attached to the DOM.

src/lib/radio/radio-group/build.json
DRiFTy17, line 1: We can delete this file

src/lib/radio/radio/build.json
DRiFTy17, line 1: This file can be deleted

src/lib/label/label-foundation.ts
```diff
@@ -55,6 +55,7 @@ export class LabelFoundation implements ILabelFoundation {
    */
   private async _initializeAsLegend(): Promise<void> {
     // Give ancestor elements a chance to connect to the DOM
+    // TODO: see if this works in Angular
```
DRiFTy17, line 58: Yeah this seems fragile

src/lib/field-next/field.test.ts
DRiFTy17, line 1: If we don't plan on including tests with this first go of changes, can you add a todo item to the board for it?

src/lib/core/base/base-component.ts
```diff
@@ -5,4 +5,9 @@ export interface IBaseComponent extends HTMLElement {
   attributeChangedCallback?(name: string, oldValue: string, newValue: string): void;
 }
 
-export abstract class BaseComponent extends HTMLElement implements IBaseComponent {}
+export abstract class BaseComponent extends HTMLElement implements IBaseComponent {
+  // We provide default implementation of these methods so that they can be extended by other
+  // classes.
+  public connectedCallback(): void {}
+  public attributeChangedCallback(name: string, oldValue: string | null, newValue: string | null): void {}
```
DRiFTy17, line 12: I thought we determined that we didn't need these at one point? Am I misremembering?

src/lib/core/styles/tokens/typography/_tokens.label.scss
```diff
@@ -26,6 +33,7 @@ $overline: utils.inherit-map(core.$base, (
 
 $tokens: (
   label: $label,
```
DRiFTy17, line 35: We should rename "label" to "label1" if we're going to introduce a new "label2" style for consistency with our type scale. This will require a sweep of usages across the library.

src/lib/date-picker/base/base-date-picker-adapter.ts
```diff
@@ -56,7 +56,7 @@ export interface IBaseDatePickerAdapter extends IBaseAdapter {
   setCalendarLocale(locale: string | undefined): void;
   propagateCalendarKey(evt: KeyboardEvent): void;
 }
-export abstract class BaseDatePickerAdapter<T extends BaseComponent> extends BaseAdapter<T> implements IBaseDatePickerAdapter {
+export abstract class BaseDatePickerAdapter<T extends IBaseComponent> extends BaseAdapter<T> implements IBaseDatePickerAdapter {
```
DRiFTy17, line 59: We only needed this changed due to the `BaseComponent` change above for `connectedCallback`. Is that change critical to this changeset?

src/lib/label/label-constants.ts
```diff
@@ -11,7 +11,7 @@ const elementName: keyof HTMLElementTagNameMap = `${COMPONENT_NAME_PREFIX}label`
 const attributes = {
   FOR: 'for',
   DYNAMIC: 'dynamic',
-  STATIC: 'static',
+  STATIC: 'static', // TODO: possibily change this to 'non-interactive'
```
DRiFTy17, line 14: I'm ok with making this change.

DRiFTy17 (commented):
See comments below, but overall things are looking really good.

We should make sure to ensure that we use the `override()` function in the Sass where possible instead of resetting tokenized styles directly.

Can you also look through the changes and remove anything that is not critical to this changeset? Seems like we have some changes to files outside of the field here.

src/lib/date-picker/base/base-date-picker-adapter.ts
DRiFTy17, line 59: Not that it should matter, but I assume we can revert this now?

src/lib/label/label.ts
```diff
@@ -99,7 +94,7 @@ export class LabelComponent extends BaseComponent implements ILabelComponent {
   public dynamic: boolean;
 
   @FoundationProperty()
-  public static: boolean;
+  public nonInteractive: boolean;
```
DRiFTy17, line 97: We probably missed this from the original PR, but can you add the `declare` keyword to all public properties in this file that use the `@FoundationProperty` decorator?

src/dev/pages/field/field.ejs
```diff
@@ -0,0 +1,67 @@
+<div class="vert">
+  <div>
+    <h3 class="forge-typography--heading2">Default</h3>
+    <forge-field>
+      <label slot="label">Label</label>
+      <input type="text" value="Content" />
+    </forge-field>
```
DRiFTy17, line 7: I noticed a small issue here when I removed the `<label>` element:
```html
<forge-field>
  <input type="text" />
</forge-field>
```

It looks like it's still applying the inset animation to the content when there is no `<label>`.

DRiFTy17 (commented):
Just a couple minor things below.

src/dev/pages/field/field.ejs
samrichardsontylertech: I almost don't think that's a bug. If someone wants to use this without a label they should set `labelPosition` to `none` 🤷‍♂️

But I'll fix it anyway.

src/lib/field-next/field-constants.ts
```diff
@@ -0,0 +1,96 @@
...
+};
+
+const animations = {
+  FLOAT_IN_LABEL: 'float-in-label-animation',
+  FLOAT_OUT_LABEL: 'float-out-label-animation'
+};
+
+const values = {
+  ANIMATION_TIMEOUT_DURATION: 1000
+};
```
DRiFTy17, line 75: It didn't look like this was used for anything?

src/lib/field-next/field-foundation.ts
```diff
@@ -0,0 +1,290 @@
...
+
+  constructor(private _adapter: IFieldAdapter) {
+    this._labelSlotListener = () => this._onSlotChange('label');
+    this._startSlotListener = () => this._onSlotChange('start');
+    this._endSlotListener = () => this._onSlotChange('end');
+    this._accessorySlotListener = () => this._onSlotChange('accessory');
+    this._supportStartSlotListener = () => this._onSlotChange('support-start');
+    this._supportEndSlotListener = () => this._onSlotChange('support-end');
+    this._popoverIconClickListener = () => this._onPopoverIconClick();
+  }
```
DRiFTy17, line 60: You could probably combine the declarations and initialization using `.bind(this, <type>)` to reduce some code:
```ts
private _labelSlotListener: () => void = this._onSlotChange.bind(this, 'label');
```

DRiFTy17 (approved):

PR #470, opened 2024-02-13
feat(divider): refactor to use tokens

jonathanearl-tylertech:
## PR Checklist

- Tests for the changes have been added/updated: y?
- Docs have been added/updated: not yet
- Does this PR introduce a breaking change? n
- I have linked any related GitHub issues to be closed when this PR is merged? y

## Describe the new behavior?
adds configuration tokens for
- border color
- border width
- border style
- margin

attributes, properties
- vertical


src/lib/core/styles/tokens/divider/_tokens.scss
```diff
@@ -0,0 +1,15 @@
+@use 'sass:map';
+@use '../../theme';
+@use '../../utils';
+@use '../../border';
+
+$tokens: (
+  color: utils.module-val(divider, color, theme.variable(outline)),
+  width: utils.module-val(divider, width, border.variable(thin)),
+  border-style: utils.module-val(divider, border-style, solid),
+  margin: utils.module-val(divider, margin, 0),
```
DRiFTy17, line 10: The more I think about this, I think we may want to just remove this `margin` token and the `margin` style from within the element.. See my other comment about the `:host-context` as well.

src/lib/divider/_core.scss
```diff
@@ -0,0 +1,37 @@
...
+  contain: content;
+}
+
+@mixin base() {
+  border: none;
+  border-bottom-color: #{token(color)};
+  border-bottom-width: #{token(width)};
+  border-bottom-style: #{token(border-style)};
+  
+  margin: #{token(margin)};
```
DRiFTy17, line 16: See other comment about removing this.

src/lib/divider/divider.scss
```diff
@@ -1,26 +1,32 @@
...
 
+.forge-divider {
+  @include base;
+}
+
+// :host-context(forge-drawer) {
+//   .forge-divider {
+//     margin: var(--forge-divider-drawer-margin, 4px) 0;
+//   }
+// }
```
DRiFTy17, line 23: @samrichardsontylertech do you have any opinions on this? I'm tempted to just remove it altogether and let developers styles it themselves. In which case, I think removing the `margin` token/style as mentioned above would be fine since developers can add `margin` themselves to the host.

To replace this we would need to go into the drawer styles and add something along the lines of the following:
```scss
::slotted(forge-divider) {
  @include divider.provide-theme(( margin: 4px ));
}
```

This won't work quite the same as `:host-context` though since it will only targets direct descendants of drawers. We could potentially use `::slotted(*)` but that may have unintended side-effects for deeply nested structures...

DRiFTy17 (commented):
Looks great!

The one other piece we need to handle aside from what is mentioned below is moving the semantics from within the shadow DOM to the host element. I'll DM you for specifics on how we handle that.

DRiFTy17:
@samrichardsontylertech in regards to the default ARIA of this element. Do you think we should use `separator` or leave it without any semantics by default?

src/lib/divider/divider.scss
samrichardsontylertech: I think it's reasonable and along the same lines as our other components to let drawers style their dividers. I'm afraid that if we take it out we're just not going to see it implemented consistently. `:host-context` seems like the *correct* way to handle this, but it doesn't have great support. `::slotted(forge-divider)` feels good too but is basically a non-starter because of how easy it is to accidentally circumvent. My gut says to go with `::slotted(*)` and if someone needs to reset the margin they can use the custom property.

That said, I think it makes a lot more sense for `margin` to be set on the host like it would for any other element. So I'd only use the internal declaration for context dependent stuff like this. Or, even better, set the token on the host so it can be easily overridden.

samrichardsontylertech:
> @samrichardsontylertech in regards to the default ARIA of this element. Do you think we should use `separator` or leave it without any semantics by default?

I'm torn on this. The closest native analogue to this is `<hr>`, which does have an implicit `separator` role. On the other hand I wouldn't be surprised if most of our dividers are used decoratively. Maybe that's the issue? It probably makes more sense to replace most of our decorative use cases with CSS borders and position the divider as a solution for separating distinct sections.

Right now I'm leaning toward defaulting to `separator` and documenting the `role` attribute as part of our API.

src/lib/divider/divider.scss
DRiFTy17: Yeah I agree. That's the direction I'm leaning as well.

@jonathanearl-tylertech I'll follow up with you on how we can implement host tokens to allow for setting the margin there instead of internally and we can use `::slotted(*)` for this from the drawer styles.

src/lib/divider/divider.ts
```diff
@@ -1,29 +1,69 @@
...
+const BaseClass = WithDefaultAria(WithElementInternals(BaseComponent));
+
 /**
- * The custom element class behind the `<forge-divider>` element.
- * 
  * @tag forge-divider
+ * 
+ * @summary Divider is used to seperate elements.
+ * 
+ * @attribute {boolean} vertical - Controls if the divider is displayed vertically or horizontally.
```
DRiFTy17, line 28: It looks like we are missing the JSDoc for the `@property`, `@csspart`, and `@slot` pieces.

src/lib/divider/divider.ts
```diff
@@ -1,29 +1,69 @@
...
-import { BaseComponent, IBaseComponent } from '../core/base/base-component';
+import { attachShadowTemplate, coerceBoolean, CustomElement } from '@tylertech/forge-core';
+import { BaseComponent } from '../core/base/base-component';
 import { DIVIDER_CONSTANTS } from './divider-constants';
+import { IWithDefaultAria, WithDefaultAria } from '../core/mixins/internals/with-default-aria';
+import { IWithElementInternals, WithElementInternals } from '../core/mixins/internals/with-element-internals';
 
 import template from './divider.html';
 import styles from './divider.scss';
+import { setDefaultAria } from '../constants';
```
DRiFTy17, line 9: Minor code style thing, but could we move this to the group of imports above? Just helps us to differentiate normal module imports from templates and styles in component files.

DRiFTy17 (approved):

Issue #477, opened 2024-02-21
[@next] create a clear button for text fields


samrichardsontylertech:
**Describe the feature:**
Create a clear button that is built into the Forge Text Field component. When clicked, the button should delete all text entered into thew field. It should not be included in the tab order and hide when no text is entered.

**Describe any alternatives you've considered:**
Consumers can create and slot in their own clear buttons. However, this is prone to differing designs, inconsistent behavior, and accessibility issues related to label and focus order.

PR #478, opened 2024-02-21
[@next][text-field] refactor to use field component

### LINKED ISSUE #476: [@next] update text field to use new field component
**Describe the feature:**
Update the text field take advantage of the new Forge Field component. Most or all styling should be handled by the field. Also update to use style tokens and follow other patterns being implemented in version 3.0.

samrichardsontylertech:
## PR Checklist

## Describe the new behavior?
The text field now uses the field component for all styling. It's now responsible only for responding to changes to the slotted input's value, disabled state, and label, and setting the same.

An optional clear button has also been added to the text field. It appears when the `showClear` property is `true` and the text has been entered into the input.

## Additional information
The field component handles multi-input styling previously handled by the text field. The date range picker, as the only component that typically uses multiple inputs, now handles inserting its own separator element.

Closes #476 and #477


src/lib/text-field/text-field.html
```diff
@@ -1,18 +1,22 @@
...
+  <forge-field id="field">
+    <slot slot="label" name="label"></slot>
+    <slot slot="start" name="start"></slot>
+    <slot></slot>
+    <slot slot="end" name="end"></slot>
+    <slot slot="end" name="clear-button">
+      <forge-icon-button aria-label="Clear" tabindex="-1" dense>
+        <forge-icon name="clear"></forge-icon>
+      </forge-icon-button>
+      <forge-tooltip>Clear</forge-tooltip>
```
DRiFTy17, line 11: I think we should add a slot here, and set the tooltip type to "label" to remove the `aria-label`. This will allow for consumers to provide their own labels for localization.
```html
<forge-icon-button tabindex="-1" dense>
  <forge-icon name="clear"></forge-icon>
</forge-icon-button>
<forge-tooltip type="label">
  <slot name="clear-tooltip">Clear</slo
</forge-tooltip>
```

src/lib/text-field/text-field.html
```diff
@@ -1,18 +1,22 @@
...
-        <slot name="addon-end"></slot>
-      </div>
-    </div>
-    <slot name="helper-text"></slot>
-  </div>
+  <forge-field id="field">
+    <slot slot="label" name="label"></slot>
+    <slot slot="start" name="start"></slot>
+    <slot></slot>
+    <slot slot="end" name="end"></slot>
```
DRiFTy17, line 6: We may want to provide the "leading" and "trailing" slots for backwards compatibility.

src/lib/text-field/text-field-adapter.ts
```diff
@@ -1,130 +1,187 @@
...
+
+  public get hasPlaceholder(): boolean {
+    return this._inputElements.some(el => !!el.placeholder);
+  }
 
   constructor(component: ITextFieldComponent) {
     super(component);
+    this._fieldElement = getShadowElement(component, TEXT_FIELD_CONSTANTS.selectors.FIELD) as IFieldComponent;
+    this._clearButtonSlotElement = getShadowElement(component, TEXT_FIELD_CONSTANTS.selectors.CLEAR_BUTTON_SLOT) as HTMLSlotElement;
+    this._fieldElement.setAttribute('exportparts', Object.values(FIELD_CONSTANTS.parts).join(', '));
```
DRiFTy17, line 41: Are adding the `exportparts` attr dynamically here to just reduce the noise in the static template?

src/lib/text-field/text-field-adapter.ts
```diff
@@ -1,130 +1,187 @@
...
+  public addRootListener(name: keyof HTMLElementEventMap, listener: EventListener): void {
+    this._fieldElement.addEventListener(name, listener);
   }
 
-  public override destroy(): void {
-    super.destroy();
-    if (this._inputMutationObserverInstances) {
-      this._inputMutationObserverInstances.forEach(mo => mo.disconnect());
+  public disableInput(disabled: boolean): void {
+    if (this._inputElements.length) {
```
DRiFTy17, line 50: This if statement isn't needed

src/lib/text-field/text-field-adapter.ts
```diff
@@ -1,130 +1,187 @@
...
 
-export class TextFieldAdapter extends FieldAdapter implements ITextFieldAdapter {
-  protected _inputElements: Array<HTMLInputElement | HTMLTextAreaElement> = [];
-  protected _inputMutationObserverInstances: MutationObserver[] = [];
+export class TextFieldAdapter extends BaseFieldAdapter implements ITextFieldAdapter {
+  protected readonly _fieldElement: IFieldComponent;
+  private readonly _clearButtonSlotElement: HTMLSlotElement;
+  private _inputElements: (HTMLInputElement | HTMLTextAreaElement)[] = [];
+  private _inputMutationObserver?: MutationObserver;
+  private _destroyValueChangerListeners: (() => void)[] = [];
```
DRiFTy17, line 27: Are we cleaning these up when the element is disconnected?

src/lib/text-field/text-field-adapter.ts
```diff
@@ -1,130 +1,187 @@
...
+    this._fieldElement.floatLabel = this.hasValue || this.hasPlaceholder;
   }
 
-  public override setInputAttributeObserver(listener: (name: string, value: string) => void): void {
-    this._applyToInputs(input => {
-      const observer = createElementAttributeObserver(input, listener, FIELD_CONSTANTS.observedInputAttributes);
-      this._inputMutationObserverInstances.push(observer);
-    });
-  }
+  public tryConnectSlottedLabel(slot: HTMLSlotElement): void {
```
DRiFTy17, line 135: When I was testing with a screen reader I noticed that the labels weren't getting attached to the `<input>` for both the native `<label>` and `<forge-label>`. I saw we have tests for this, but wasn't sure if I was missing something.

src/lib/text-field/text-field-foundation.ts
```diff
@@ -1,15 +1,110 @@
...
+        this.disabled = value !== null;
+        break;
+      case 'placeholder':
+        this._tryFloatLabel();
+        break;
+    }
+  }
+
+  private _onClearButtonClick(): void {
+    const allowClear = this._adapter.emitHostEvent(TEXT_FIELD_CONSTANTS.events.CLEAR, undefined, true, true);
```
DRiFTy17, line 54: Let's use the `dispatchHostEvent()` method here instead of `emitHostEvent()`. We should mark `emitHostEvent()` as deprecated as well.

src/lib/date-picker/base/base-date-picker-adapter.ts
```diff
@@ -1,12 +1,14 @@
 import { getShadowElement, randomChars } from '@tylertech/forge-core';
+import { get } from 'https';
```
DRiFTy17, line 2: Remove this `import`

src/lib/field-next/base/base-field.test.ts
DRiFTy17, line 1: Are you planning to add tests here now? If not, let's create a todo item on the board for this task.

src/dev/pages/text-field/text-field.ejs
```diff
@@ -1,21 +1,36 @@
...
 
-  <!-- Label and input -->
-  <label for="text-field-input">First name</label>
-  <input type="text" id="text-field-input" />
+  <div>
+    <h3 class="forge-typography--heading2">Multiple inputs</h3>
+    <forge-text-field>
+      <label>First and last name</label>
+      <input type="text" />
+      <input type="text" />
```
DRiFTy17, line 23: For multiple inputs in text fields, I wonder if we should manage the tabindex of the second input to only be keyboard accessible if the first input has a value? We could add a property/attribue to control whether that happens as well? Additionally, if you attempt to click into that second input in this scenario, it could also immediately move focus to the first input as well.

Maybe add a todo item to the board for this so we can revisit if we don't want to hold this up right now.

DRiFTy17 (commented):
Looks good! Visually everything seems to work great in the demos. See comments below about accessibility/screen reader issues I ran into.

Looks like some build issues to be addressed as well, and we need to make sure this component is 100% backwards compatible.

src/lib/text-field/text-field-adapter.ts
samrichardsontylertech: Mostly so we can define the parts just in the field constants and use them in every component that includes a field. If it weren't for that I'd prefer to add them statically, but this ends up being DRYer.

src/lib/text-field/text-field-adapter.ts
samrichardsontylertech: Yeah, we should do that

src/dev/pages/text-field/text-field.ejs
samrichardsontylertech: Let's split this out

src/lib/field-next/base/base-field.test.ts
samrichardsontylertech: I'll address this later

src/lib/text-field/text-field-adapter.ts
samrichardsontylertech: Looking into it, the problem seems to be that when the example loads there is an `<input>` and a `<textarea>` slotted. Because there are two elements the component doesn't try to label either of them.

DRiFTy17 (approved):

PR #482, opened 2024-02-22
Next stack refactor

nickonometry:
## PR Checklist

- I have linked any related GitHub issues to be closed when this PR is merged? [Y/N]

## Describe the new behavior?
Converted stack component to use design tokens. New tokens include:
- gap
- height
- stretch
- alignment
- justify
- wrap

## Additional information
A request has come up several times to allow for vertical alignment of child items when used inline. One specific situation this was noticeable was putting two items in an inline stack that have different heights or have margin-block/padding-block applied. To resolve this, the alignment property has been updated to map to the CSS align-items property, and a new justify property has been added to control justify-content. I also added examples/demos to the EJS files to replicate the common scenarios where issues were discovered. The stack should handle all intended use cases now and mimics figmas autolayout functionality 1 to 1


src/dev/pages/stack/stack.scss
```diff
@@ -3,12 +3,36 @@
 }
 
 .box {
-  padding: 36px;
   border: 4px dashed var(--mdc-theme-text-secondary-on-background);
   background-color: var(--mdc-theme-text-disabled-on-background);
```
DRiFTy17, line 7: Let's replace any `--mdc-*` variable in these demos with the new `--forge-*` equivalents so we don't have to do it later when we fully remove MDC.

src/dev/pages/stack/stack.ejs
```diff
@@ -53,20 +155,40 @@
...
-              <input type="text" id="input-text-01" />
-              <label for="input-text-01" slot="label">Text field</label>
-            </forge-text-field>
-            <forge-text-field style="--forge-stack-stretch: 4">
-              <input type="text" id="input-text-01" />
-              <label for="input-text-01" slot="label">Text field</label>
-            </forge-text-field>
-            <forge-text-field style="--forge-stack-stretch: 6">
+        <forge-stack>
+          <forge-stack gap="16">
```
DRiFTy17, line 159: We should remove the `gap` attribute when it matches the default value. Just makes it easier when we share these demos out so people don't think they _need_ it.

DRiFTy17 (approved):

PR #487, opened 2024-02-28
feat(popover): new delay property

nickonometry:
## PR Checklist

## Describe the new behavior?
When the trigger type of the popover was set to hover, it felt like the popover was appearing too fast. Adding the delay property keeps it consistent with the tooltip, and allows us to control the speed at which the popover element appears

### DIFF
--- a/src/dev/pages/popover/popover.html
+++ b/src/dev/pages/popover/popover.html
@@ -4,6 +4,7 @@ include('./src/partials/page.ejs', {
     title: 'Popover',
     includePath: './pages/popover/popover.ejs',
     options: [
+      { type: 'text-field', inputType: 'number', id: 'opt-delay', label: 'Delay', defaultValue: 0 },
       {
         type: 'select',
         label: 'Placement',
--- a/src/dev/pages/popover/popover.html
+++ b/src/dev/pages/popover/popover.html
@@ -4,7 +4,7 @@ include('./src/partials/page.ejs', {
     title: 'Popover',
     includePath: './pages/popover/popover.ejs',
     options: [
-      { type: 'text-field', inputType: 'number', id: 'opt-delay', label: 'Delay', defaultValue: 0 },
+      { type: 'text-field', inputType: 'number', id: 'opt-hover-delay', label: 'Hover Delay', defaultValue: 0 },
       {
         type: 'select',
         label: 'Placement',
--- a/src/lib/popover/popover-foundation.ts
+++ b/src/lib/popover/popover-foundation.ts
@@ -13,7 +13,7 @@ export interface IPopoverFoundation extends IOverlayAwareFoundation {
   longpressDelay: number;
   persistentHover: boolean;
   hoverDismissDelay: number;
-  delay: number;
+  hoverDelay: number;
   dispatchBeforeToggleEvent(state: IDismissibleStackState): boolean;
 }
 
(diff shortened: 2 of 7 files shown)


src/lib/tooltip/tooltip.test.ts
```diff
@@ -434,6 +434,17 @@ describe('Tooltip', () => {
       expect(harness.isOpen).to.be.true;
     });
 
+    it('should open when hovering the trigger button and a delay is set', async () => {
+     const harness = await createFixture({ triggerType: 'hover', delay: 500 });
```
DRiFTy17, line 438: Did you intend to add this test to the `popover.test.ts` file instead?

src/lib/popover/popover-foundation.ts
```diff
@@ -304,7 +307,13 @@ export class PopoverFoundation extends BaseClass implements IPopoverFoundation {
       if (!this._persistentHover) {
         this._adapter.addAnchorListener('mouseleave', this._anchorMouseleaveListener);
       }
-      this._openPopover();
+      if (this._delay) {
+        this._hoverTimeout = window.setTimeout(() => {
+        this._openPopover();
+       }, this._delay);
```
DRiFTy17: Can you fix the indentation on these two lines?

src/lib/popover/popover-foundation.ts
```diff
@@ -526,6 +536,16 @@ export class PopoverFoundation extends BaseClass implements IPopoverFoundation {
     }
   }
 
+  public get delay(): number {
+    return this._delay;
+  }
+  public set delay(value: number) {
+    if (this._delay !== value) {
+      this._delay = value;
```
DRiFTy17: Let's add some safe guards for `NaN` and negative numbers here and just use the default hover delay if detected. May be worth updating the tooltip `delay` property to guard against that the same way as well.

```typescript
public set delay(value: number) {
  if (isNaN(value) || value < 0) {
    value = POPOVER_CONSTANTS.defaults.hoverDelay;
  }
  if (this._delay !== value) {
    // ...
  }
}
```

We should add a test that captures trying to set these values as well.

src/dev/pages/popover/popover.html
```diff
@@ -4,6 +4,7 @@
     title: 'Popover',
     includePath: './pages/popover/popover.ejs',
     options: [
+      { type: 'text-field', inputType: 'number', id: 'opt-delay', label: 'Delay', defaultValue: 0 },
```
DRiFTy17, line 7: Let's rename the `label` to "Hover delay", and `id` to "opt-hover-delay" for consistency, and update the demo `.ts` file as well.

src/lib/tooltip/tooltip.test.ts
nickonometry: I was referencing the tooltip file and accidentally added it there. I can remove it from there and move it over to the popover

src/lib/tooltip/tooltip.test.ts
```diff
@@ -434,6 +434,17 @@ describe('Tooltip', () => {
       expect(harness.isOpen).to.be.true;
     });
 
+    it('should open when hovering the trigger button and a delay is set', async () => {
```
DRiFTy17, line 437: You can remove this test since it's a duplicate

src/lib/popover/popover.test.ts
```diff
@@ -1428,6 +1450,7 @@ interface IPopoverFixtureConfig {
   animationType?: PopoverAnimationType;
   triggerType?: PopoverTriggerType;
   persistentHover?: boolean;
+  hoverDelay?: any;
```
DRiFTy17, line 1453: Use `number` here instead and for your NaN test just set `hoverDelay: 'Testing' as any` to force override the type. We want strong typing for everything else on this property.

src/lib/popover/popover.ts
```diff
@@ -138,6 +142,9 @@ export class PopoverComponent extends OverlayAware<IPopoverFoundation> implement
       case POPOVER_CONSTANTS.attributes.PERSISTENT_HOVER:
         this.persistentHover = coerceBoolean(newValue);
         return;
+      case POPOVER_CONSTANTS.attributes.HOVER_DELAY:
+        this.delay = coerceNumber(newValue);
```
DRiFTy17, line 146: You forgot to rename this property

DRiFTy17 (dismissed):

DRiFTy17 (approved):

PR #489, opened 2024-02-28
[@next][button-area] refactor to use tokens, state-layer, and focus-indicator

derekmoss:
## PR Checklist

- Tests for the changes have been added/updated:Y

## Describe the new behavior?

adds configuration tokens for

    primary-color
    cursor
    disabled cursor
    focus indicator color
    focus indicator offset

adds state layer

adds focus indicator


src/lib/button-area/_core.scss
```diff
@@ -0,0 +1,25 @@
...
+  display: block;
+  position: relative;
+}
+
+@mixin base {
+  position: relative;
+
+  outline: none;
+  user-select: none;
+  -webkit-appearance: none;
```
samrichardsontylertech, line 15: No need to set `outline` and `appearance` here. The root element is just a `div` so the typical `button` overrides don't do anything. Also remove `user-select` — the button area is intended to wrap complex content that the user might want to select. There's code in the component to prevent a click if a selection was made.

samrichardsontylertech (changes requested):
Looks great overall! Just remove a few styles and I'll give it an approval

src/lib/button-area/button-area.scss
```diff
@@ -1,11 +1,62 @@
...
+    }
+  }
+
+  &.forge-button-area--disabled {
+    @include disabled;
+  }
+
+  &__button {
+    @include utils.visually-hidden;
+  }
```
DRiFTy17, line 47: To build on my previous comments, we can probably simplify and remove nesting for these selectors once you remove BEM and use host attributes for things like `disabled`.

src/lib/button-area/button-area.scss
```diff
@@ -1,11 +1,62 @@
...
+@use '../state-layer';
+@use '../focus-indicator';
+@use '../utils/mixins' as utils;
 
-@include mixins.core-styles;
+//
+// Host
+//
+
+$host-tokens: [disabled-cursor focus-indicator-color];
```
DRiFTy17, line 10: Do these _need_ to be host-level tokens? I think you can just keep them defined within the shadow root.

src/lib/button-area/button-area.ts
```diff
@@ -19,13 +20,33 @@ declare global {
 
 /**
  * The custom element class behind the `<forge-button-area>` element.
```
DRiFTy17, line 22: Let's get rid of this text. I have been removing it as I go since it's not adding any value.

src/lib/button-area/button-area.test.ts
DRiFTy17, line 1: It looks like you have some uncovered code when looking at the code coverage. We may be missing a few more tests.

DRiFTy17 (commented):
Things are looking good! From a visual and interaction perspective I think it's great. Just some minor things below to review.

I asked @samrichardsontylertech to take a look as well since he had some plans for this component in the past.

DRiFTy17 (dismissed):

samrichardsontylertech (dismissed):

DRiFTy17 (approved):

PR #490, opened 2024-03-01
Next page state

tyl-paul-lapczynski:
## PR Checklist

## Describe the new behavior?
Tokenize the forge-page-state. Changed h1 and h2 elements to divs. Will rely on devs to make that choice. Changed to use @container queries.


src/lib/page-state/_core.scss
```diff
@@ -0,0 +1,84 @@
...
+  justify-content: center;
+  grid-auto-flow: column;
+  gap: #{token(actions-spacing)};
+}
+
+@mixin actions-mobile {
+  display: flex;
+  justify-content: center;
+  flex-direction: column;
+  align-items: center;
```
DRiFTy17, line 82: We can remove all of these styles. The only important one is `grid-auto-flow: row` now that the `.actions` container element is using a grid layout.

src/lib/page-state/_core.scss
```diff
@@ -0,0 +1,84 @@
...
+@mixin graphic-mobile {
+  @include override(height, mobile-graphic-height);
+}
+
+@mixin title {
+  @include typography.style(heading8);
+
+  color: #{token(title-color)};
+  margin-top: #{token(title-spacing)};
+  margin-bottom: #{token(title-spacing)};
```
DRiFTy17, line 53: Use `margin-block: #{token(title-spacing)};` here instead

src/lib/page-state/_core.scss
```diff
@@ -0,0 +1,84 @@
...
+
+@mixin title-mobile {
+  @include typography.style(heading5);
+}
+
+@mixin message {
+  @include typography.style(body4);
+
+  color: #{token(message-color)};
+  margin-bottom: #{token(message-spacing)};
```
DRiFTy17, line 64: Use `margin-block-end: #{token(message-spacing)};` here instead

src/lib/utils/_mixins-container.scss
DRiFTy17, line 1: I added an issue to our board to update `stylelint` to fix the linting issue related to `@container` in this file. It requires our node version to be updated to a minimum of v18 first, which I'm planning to do shortly as well so I'll take care of it then.

src/lib/page-state/_core.scss
```diff
@@ -0,0 +1,84 @@
...
+  margin-bottom: #{token(spacing)};
+
+  text-align: center;
+}
+
+@mixin base-mobile {
+  @include override(width, mobile-width);
+}
+
+@mixin graphic {
```
DRiFTy17, line 29: Can I suggest we rename the `graphic` mixin to `graphic-container` and rename `graphic-slotted` to `graphic`? I think this would better reflect what these mixins are intended to be applied to. The class names in the template are fine as-is IMO.

src/lib/page-state/_core.scss
```diff
@@ -0,0 +1,84 @@
...
+@mixin host {
+  display: flex;
+  width: 100%;
+  justify-content: center;
+  container-type: inline-size;
+}
+
+@mixin host-hidden {
+  display: none;
+}
```
DRiFTy17, line 15: Super minor, and I know I'm contradicting myself a little when I say this, but you can just revert this back to putting the individual style back in the `page-state.scss` file. I only say to do this because every custom element style has the same thing defined to mimic native HTML support, and it will never need to be a mixin that we intend to reuse. I don't mind in the end, but just wanted to share it with you as you go forward with other components so you know not worry about this special case selector.

samrichardsontylertech (approved):

DRiFTy17 (approved):

PR #496, opened 2024-03-12
feat(stepper): refactor to use tokens, focus-indicator, and state-layer

tyl-paul-lapczynski:
## PR Checklist

## Additional information
Ignoring Tests for now. Will add a separate ticket.


src/lib/core/styles/tokens/stepper/stepper/_tokens.scss
DRiFTy17, line 1: If we don't need tokens for the stepper, let's go ahead and remove this file.

src/lib/stepper/step/_core.scss
```diff
@@ -0,0 +1,377 @@
...
+
+@forward './token-utils';
+
+@mixin base() {
+  -webkit-tap-highlight-color: transparent;
+  padding: 12px 16px;
+  outline: none;
+  background: none;
+  border: none;
+  border-radius: 14px;
```
DRiFTy17, line 14: Let's add a shape token here, and use either `large` or `extra-large`

src/lib/stepper/step/_core.scss
```diff
@@ -0,0 +1,377 @@
...
+
+  border-style: solid;
+  border-width: 2px;
+  border-color: transparent;
+  border-radius: 50%;
+
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  transition: background-color 400ms cubic-bezier(0.25, 0.8, 0.25, 1);
```
DRiFTy17, line 72: Let's use animation tokens for the duration and easing

src/lib/stepper/stepper/_token-utils.scss
DRiFTy17, line 1: Let's remove this file since the stepper doesn't use tokens

src/lib/stepper/step/_core.scss
```diff
@@ -0,0 +1,377 @@
...
+  color: #{token(error-color)};
+}
+
+@mixin line-clustered {
+  @include override(line-min-width, line-min-width-clustered);
+}
+
+@mixin vertical-base {
+  width: 100%;
+  border-radius: 4px;
```
DRiFTy17, line 183: We need to propagate this `border-radius` to the focus indicator when it's vertical. Once the base `border-radius` is using a shape token, you can just override it here instead and then it would automatically propagate to the focus indicator.

DRiFTy17 (commented):
Everything looks great. Just a few minor comments below that I think we may want to include now.

We were planning to rewrite this component for 3.0 originally, but I don't think it will make it into this release so we'll take care of that at a later time and simplify things.

DRiFTy17 (approved):

DRiFTy17 (approved):

PR #500, opened 2024-03-18
feat(list-dropdown): refactored to use `<forge-popover>` instead of `<forge-popup>` internally

DRiFTy17:
Refactors the select, menu, autocomplete, and time-picker to use the new `<forge-popover>` through by way of the list-dropdown refactor.

Additionally, some changes were made to the popover:

1. Added exit animations, and updated dependent logic and tests to account for the timeout between the exit animation and the element being removed from the DOM.
2. Added a new `preset="dropdown"` attribute to allow for setting strict pre-defined popover styles for use as dropdowns.
3. Added tokens for various popover surface styles (for use with the presets or custom overrides). Typically we will expect the provided content to drive the design however, so presets and tokens are more of an advanced usage or internal concept.
4. Added support for the legacy `constrain-viewport-width` attribute that was used on the `<forge-popup>` to ensure the popover surface could not be larger than the viewport width when used on small viewports.

### DIFF
--- a/src/lib/core/styles/tokens/popover/_tokens.scss
+++ b/src/lib/core/styles/tokens/popover/_tokens.scss
@@ -34,23 +34,24 @@ $tokens: (
   arrow-border-width: utils.module-val(popover, arrow-border-width, border.variable(thin)),
 
   // Animation
-  animation-timing: utils.module-val(popover, animation-timing, animation.variable(easing-decelerate)),
+  animation-enter-duration: utils.module-val(popover, animation-enter-duration, animation.variable(duration-short3)),
+  animation-enter-timing: utils.module-val(popover, animation-enter-timing, animation.variable(easing-decelerate)),
   animation-exit-duration: utils.module-val(popover, animation-exit-duration, animation.variable(duration-short2)),
   animation-exit-timing: utils.module-val(popover, animation-exit-timing, animation.variable(easing-accelerate)),
 
-  zoom-duration: utils.module-val(popover, zoom-duration, animation.variable(duration-short3)),
-  zoom-timing: utils.module-ref(popover, zoom-timing, animating-timing),
+  zoom-enter-duration: utils.module-ref(popover, zoom-enter-duration, animation-enter-duration),
+  zoom-enter-timing: utils.module-ref(popover, zoom-enter-timing, animation-enter-timing),
   zoom-exit-duration: utils.module-ref(popover, zoom-exit-duration, animation-exit-duration),
   zoom-exit-timing: utils.module-ref(popover, zoom-exit-timing, animation-exit-timing),
 
-  slide-duration: utils.module-val(popover, slide-duration, animation.variable(duration-short3)),
-  slide-timing: utils.module-ref(popover, slide-timing, animating-timing),
-  slide-offset: utils.module-val(popover, slide-offset, 24px),
+  slide-enter-duration: utils.module-ref(popover, slide-enter-duration, animation-enter-duration),
+  slide-enter-timing: utils.module-ref(popover, slide-enter-timing, animation-enter-timing),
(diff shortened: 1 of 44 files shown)


src/lib/core/styles/tokens/popover/_tokens.scss
```diff
@@ -28,22 +35,34 @@ $tokens: (
 
   // Animation
   animation-timing: utils.module-val(popover, animation-timing, animation.variable(easing-decelerate)),
```
samrichardsontylertech, line 37: Should we name the animation tokens so it's clear which transitions they apply to? For example, does `zoom-duration` affect entrance *and* exit durations or just entrance durations? Does `zoom-exit-timing` inherit from it? Why isn't there a `zoom-entrance-duration`? It's easy to find out by looking at this file but typical users might be confused

src/lib/menu/menu-foundation.ts
```diff
@@ -16,27 +17,27 @@ export interface IMenuFoundation extends ICustomElementFoundation {
   optionsFactory: MenuOptionFactory | undefined;
   selectedIndex: number;
   selectedValue: any;
-  placement: PopupPlacement;
-  fallbackPlacements: PopupPlacement[];
+  placement: PositionPlacement;
```
samrichardsontylertech, line 20: Maybe some of these common popover properties (placement, offset) could be made into a mixin?

src/lib/menu/menu-foundation.ts
```diff
@@ -16,27 +17,27 @@ export interface IMenuFoundation extends ICustomElementFoundation {
...
+  fallbackPlacements: PositionPlacement[];
   dense: boolean;
   iconClass: string;
   persistSelection: boolean;
   mode: MenuMode;
-  popupOffset: IPopupPosition;
+  popupOffset: IOverlayOffset;
```
samrichardsontylertech, line 26: Should we alias this as `popoverOffset`? The same goes for any other components with similar properties

src/lib/popover/_animations.scss
```diff
@@ -20,6 +30,16 @@
   }
 }
 
+@keyframes slideout {
```
samrichardsontylertech, line 33: The animations aren't respecting prefers reduced motion. Not a big deal for fade in/out, but sliding and zooming should automatically disable.

samrichardsontylertech:
Are you planning on replacing the popup in the calendar dropdown? Otherwise I think this pretty much covers everything

src/lib/menu/menu-foundation.ts
DRiFTy17: I thought about this, but since we intend to deprecate these components in favor of new patterns (inline/static usage) I figured I would save that until then and keep these changes focused on just list-dropdown.

src/lib/menu/menu-foundation.ts
DRiFTy17: I think semantically they could still work as "popup" for now, and it will avoid confusion since we plan to deprecate these at some point soon. Eventually with the new patterns in place we will be setting these through the dropdown list element(s) anyway, and in which case I think we'll be using more generic names like just "offset" for those.

DRiFTy17:
> Are you planning on replacing the popup in the calendar dropdown? Otherwise I think this pretty much covers everything

Yes, that will be done in a separate PR. These changes only pertain to components that use list dropdown.

src/lib/core/styles/tokens/popover/_tokens.scss
DRiFTy17: Yeah good call. I'll rename those so they specify "enter" in the name of the token now that we have both enter and exit tokens.

samrichardsontylertech (approved):

PR #512, opened 2024-03-28
feat(forge-core): update core styles

DRiFTy17:
This change adjusts the `forge-core.css` stylesheet by adjusting our default styles to be more flexible, without overriding `overflow` to better support mobile usage.

- Use dynamic viewport units for `height` and `width` on the `<body>`
- Remove `overflow: hidden` on the `<body>`
- Remove applying styles on the root `<html>` element

### DIFF
--- a/src/lib/forge-core.scss
+++ b/src/lib/forge-core.scss
@@ -1,9 +1,5 @@
-@use './utils/mixins-core';
-
-body, html {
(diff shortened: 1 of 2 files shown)


samrichardsontylertech (approved):
Looks good, I'd just keep an eye on the potential performance issues of using dynamic viewport units on the body

PR #518, opened 2024-04-03
feat(chip-field): refactor to use `<forge-field>`

DRiFTy17:
- Refactor to use `<forge-field>` internally
- Update styles, patterns, and logic
- Migrate tests to WTR
- Fix small bugs in field styles


samrichardsontylertech (changes requested):
This looks good, though I found two small issues with it. When you switch to the extra-small density with an inset label a little space remains above the chips. Also, changing the theme should also change the them of the chips.

samrichardsontylertech (approved):

PR #563, opened 2024-05-14
feat(base-button): remove internal `<a>` support and APIs and instead allow for slotted `<a>` elements and update each button-like element to support slotted anchor styling/functionality

DRiFTy17:
- Removes anchor-related APIs from the base button functionality such as `href`, `target`, `rel`, `download`, and `anchor` properties.
- Updates base button logic to instead detect slotted `<a>` elements
- Slotted `<a>` do **not** work work with start/end slots, the `<a>` must be the only child, and the button will apply a flex layout to the slotted `<a>`
- Update semantics and interactions when slotting anchors
- Removed globally applied `<a>` styles in favor of opt-in classes

### DIFF
--- a/src/dev/pages/button/button.ejs
+++ b/src/dev/pages/button/button.ejs
@@ -32,37 +32,37 @@
     <div class="vert">
       <forge-button>
         <a href="javascript: console.log('Anchor button clicked');">
-          Text button
+          Anchor button
           <forge-icon name="open_in_new"></forge-icon>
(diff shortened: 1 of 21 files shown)


samrichardsontylertech (changes requested):
The anchor text in the filled and raised variants is the same color as the background. Fix that and this looks great.

samrichardsontylertech (approved):

PR #564, opened 2024-05-16
feat(autocomplete): add popoverExpanded prop/attr to base field and update autocomplete to use it

DRiFTy17:


samrichardsontylertech (changes requested):
I can change this in my branch, but you should also remove `popoverExpanded` from the field constants and pull it from here

samrichardsontylertech (approved):

PR #619, opened 2024-06-07
chore: add code formatting with Prettier

DRiFTy17:


.prettierrc
MikeMatusz, line 1: For compare/contrast purposes, [here's the prettier config](https://github.com/tyler-technologies/nwps-design-framework/blob/chore/artifactory-password/packages/prettier-config/index.json) we use for PSDF.

Out of the differences, I'm mainly wondering about the following:
- We went opposite directions on `arrowParens`
- We enabled `htmlWhitespaceSensitivity` because we saw some visual quirks when the formatting was applied to existing templates.  I think ideally we wouldn't override it, and Forge might not have relied on it anywhere, but might be worth some extra eyes going over each component quickly looking for issues.
-  Probably should set the typescript parser
- thoughts on `"semi": true`?

Ross-Blakeney (approved):
I've become a fan of prettier's simplicity. I like it!

MikeMatusz (approved):

.prettierrc
```diff
@@ -0,0 +1,8 @@
+{
+  "arrowParens": "avoid",
+  "bracketSpacing": true,
+  "bracketSameLine": true,
+  "printWidth": 160,
+  "singleQuote": true,
+  "trailingComma": "none"
```
samrichardsontylertech, line 7: Controversial opinion, but I like having trailing commas. If nothing else it's good for keeping diffs concise.

.prettierrc
MikeMatusz: On further review, `arrowParens` setting is more consistent with existing patterns, makes sense to minimize changes.  Pretty much all HTML file changes are showing as whitespace only, so it's unlikely much whitespace in changed in the middle of lines.  I think it mostly had to do with spacing around Angular bindings where we saw a few issues.  Setting the TS parser is unnecessary,  going to remove it from ours (though I see historically why it was added).  And semi probably wouldn't really make a difference.

samrichardsontylertech (approved):

.prettierrc
Ross-Blakeney: I like them as well. We can be both cause controversy together!

MikeMatusz: I wouldn't be opposed.  I instinctively hate it, but the advantage on diffs is probably enough to allow my brain to eventually rewire itself and get used to it.  Would definitely result in a lot more file changes in this repo, though.

.prettierrc
Ross-Blakeney: - `arrowParens` I set to `avoid` just because it's closer to my personal style and leads to less clutter, imo.
- I'm waffling on `htmlWhitespaceSensitivity`... I hate when the closing angle bracket get wrapped though. I use the `css` value and I don't hate it, but I do clean up wrapped brackets myself.
- `semi` defaults to true. Def my preference for it to be true.

.prettierrc
DRiFTy17: Yeah I'm fine with enabling this if everyone wants to. I agree with @MikeMatusz, I'm not a fan because the noise it adds to the files and it makes things feel incomplete to me but the diff reason is a good point. I think we should try it without them for now, and if we feel the need to enable it later we can do so.

PR #626, opened 2024-06-11
fix(tab-bar): move ARIA attributes to internal scroll container for tab bar

DRiFTy17:
## PR Checklist

## Describe the new behavior?
Moves the `role="tablist"` from the `<forge-tab-bar>` to the internal scroll container element that `<forge-tab>` elements slot into. This keeps the scroll buttons outside of the tablist element.

## Additional information
This will fail an audit with Arc, but it's due to a bug in their audit.


samrichardsontylertech (changes requested):
Looks good, I just have one thought. Maybe we should also add a `label` property in the interest of accessibility. It would just set `aria-label` on the internal element.

samrichardsontylertech (approved):

PR #631, opened 2024-06-21
fix(stepper): fixed nested interactive element accessibility issue

nickonometry:
## PR Checklist

## Describe the new behavior?
Fixed an A11y issue where focusable elements were found within an ARIA role that defines its children as presentational

### DIFF
--- a/src/lib/stepper/step/_mixins.scss
+++ b/src/lib/stepper/step/_mixins.scss
@@ -8,7 +8,6 @@
 @mixin core-styles() {
   .forge-step {
     @include base();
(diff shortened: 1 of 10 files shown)


DRiFTy17 (approved):

samrichardsontylertech (approved):
Looks good. The vertical step example with nested inputs and buttons is invalid, but we can just discourage usage like that. In the future we should add a target property to the step component to make it easy to shift focus to its associated `tabpanel` and conform to the a11y guidance a little better.

PR #634, opened 2024-06-26
docs: improve API docs, content, and demos

DRiFTy17:


src/lib/card/card.ts
```diff
@@ -23,7 +23,7 @@ declare global {
  * @property {boolean} [raised=false] - Whether the card has elevation or not.
  *
  * @attribute {boolean} [raised=false] - Whether the card has an raised or not.
```
samrichardsontylertech, line 25: Typo here — "has an raised"

samrichardsontylertech: The description also differs from the property, they should probably match

samrichardsontylertech (changes requested):
Fix that typo and this looks good 👍

PR #649, opened 2024-07-09
fix(select): temp fix to remove disabled property from form associated mixin for TS2611 until TypeScript fixes accessor union bug

DRiFTy17:
## PR Checklist

## Describe the new behavior?
This is a temporary "fix" to just omit the `disabled` property in the `WithFormAssociation` mixin until TypeScript can fix this bug in how getter accessor properties are unioned.

For now, any classes that use the `WithFormAssociation` mixin will not be required to implement the `disabled` property, but it's likely that all instance will always have that property anyway.

Here is the TypeScript issue for tracking status:
https://github.com/microsoft/TypeScript/issues/58790

### DIFF
--- a/src/dev/pages/table/table.ts
+++ b/src/dev/pages/table/table.ts
@@ -3,7 +3,6 @@ import '@tylertech/forge/table';
 import '@tylertech/forge/table/forge-table.scss';
 import './table.scss';
 import { TextFieldComponentDelegate } from '@tylertech/forge/text-field';
-import { PositionPlacement } from '@tylertech/forge-core';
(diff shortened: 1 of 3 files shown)


src/dev/pages/table/table.ts
```diff
@@ -288,7 +287,7 @@ function getMenuColumnTemplate(rowIndex: number): HTMLElement {
   button.appendChild(icon);
 
   const menu = document.createElement('forge-menu');
-  menu.placement = 'bottom-right' as PositionPlacement;
+  menu.placement = 'bottom-end';
```
samrichardsontylertech, line 290: 👍

samrichardsontylertech (approved):

PR #667, opened 2024-08-09
fix(button-toggle): relax the generic type on the event data interfaces to `any`

DRiFTy17:
## PR Checklist

## Describe the new behavior?
While I would prefer to keep `unknown` since that is technically more correct, `any` is still appropriate, and it helps alleviate some friction in framework bindings such as in Angular where you cannot access the value inline due to type conflicts. This requires the use of `$any()` in some cases.


samrichardsontylertech (approved):
Painful, but approved

PR #698, opened 2024-09-25
fix(list-item): fixed a bug where focus was being forced to propagate to the interactive element when used within a list dropdown

DRiFTy17:
## PR Checklist

- I have linked any related GitHub issues to be closed when this PR is merged?N

## Describe the new behavior?
Adds new `allowFocusPropagation` configuration to the `<forge-list-item>` element to allow for controlling whether focus is propagated to the interactive element or not (`true` by default).

This is used within the list-dropdown to ensure that when the user interactive with a non-interactive element in a list item within the dropdown (which should never receive focus) that the interactive element doesn't inadvertently get focused. When used statically within a page, list items should always propagate focus.

### DIFF
--- a/src/dev/pages/list/list.scss
+++ b/src/dev/pages/list/list.scss
@@ -2,6 +2,10 @@
   display: grid;
   grid-template-columns: 1fr 1fr;
   gap: var(--forge-spacing-large);
+
+  & > div {
+    min-width: 0;
+  }
 }
 
 .list-demo {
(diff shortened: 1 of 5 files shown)


alienriver49 (approved):
I have confirmed this fixes the issue when in place in our application, thank you for jumping on this so quickly!

samrichardsontylertech (approved):

PR #701, opened 2024-10-01
feat(calendar): enable UI text customization

samrichardsontylertech:
## PR Checklist

## Describe the new behavior?
Added the ability to customize all UI text in the calendar component via slots and through the date picker components via a `calendarText` property.

### DIFF
--- a/src/lib/calendar/_mixins.scss
+++ b/src/lib/calendar/_mixins.scss
@@ -130,10 +130,6 @@
         @include arrow-rtl;
       }
     }
-
-    slot {
-      display: none;
-    }
   }
 
   #accessible-header {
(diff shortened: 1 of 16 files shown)


MikeMatusz:
I like this approach.  It's similar to what we're doing in the PSDF, with the "IComponentText" interface and default values (though we don't really have our localization strategy fully fleshed-out yet, so take my opinion with a grain of salt, lol), but I also like how exposing the slots allows for localization via the template, since I know some teams use Angular pipes for getting localized strings.

I know there's been some start towards being able to specify "global" defaults, and I think this would benefit from being able to be set at that level, rather than having to provide it to each individual instance.  Could definitely be a future enhancement if it makes sense.

samrichardsontylertech:
@MikeMatusz Yeah, this is definitely something we can and should flesh out in the future. I'm envisioning maybe a global object of default strings that we can override on a per app basis. This is just a band aid fix though. Thankfully we don't have a lot of text across our components.

DRiFTy17 (approved):

PR #709, opened 2024-10-18
Fix tab disabled

### LINKED ISSUE #707: [tab-bar] individual tab disabled state overwritten by tab bar
**Describe the bug:**
When creating a tab group, with one tab initially disabled, as follows, the `forge-tab-bar` forces the `disabled` property of all child tabs to `false` to match its own disabled property, which causes all the tabs to be enabled.
```html
<forge-tab-bar data-aria-label="Demo tabs" active-tab="0">
   <forge-tab>Tab 1 </forge-tab>
   <forge-tab disabled>Tab 2 </forge-tab>
   <forge-tab>Tab 3 </forge-tab>
</forge-tab-bar>
```

In addition to initially overwriting the individual tab `disabled` states, the individual tab states will be lost whenever the `disabled` state on the parent tab-bar element is toggled.  So if in the above example, the tab-bar was disabled, then re-enabled, it would be impossible for any single tab to remain disabled.

**To Reproduce:**
1. Render the above snippet.
2. See that Tab 2 is not disabled, despite the attribute.
3. Set the tab disabled again and see that it does disable this time.
4. Disable, then re-enable the tab bar.
5. See that Tab 2 has become enabled again.

**Expected behavior:**
Tabs should maintain their own independent disabled state.  While it should not be possible for any tab to display as enabled when the parent tab bar is disabled, enabling the tab bar should not force all tabs to enabled, as there may still be a condition where a certain tab is not relevant.

**Screenshots:**

 - Forge version: v3.2.0
 - Browser: All
 - Platform: All
 - OS: All

**Additional context:**
[Teams Thread](https://teams.microsoft.com/l/message/19:1gCw-wLjrS56X_iAY0mpHpk4PyWOi08JiMM7Kc1xvho1@thread.tacv2/1729105258525?tenantId=7cc5f0f9-ee5b-4106-a62d-1b9f7be46118&groupId=2bb1b8ea-cd12-4df5-a4ab-0c895bfcda5c&parentMessageId=1729105258525&teamName=Forge&channelName=General&createdTime=1729105258525)

MikeMatusz:
Fixes #707 

## PR Checklist

- Docs have been added/updated: n/a

## Describe the new behavior? 
Open to any suggestions on a more idiomatic way to accomplish this.  Was originally going to keep the `forge-tab--disabled` class abstracted inside the Shadow DOM, but currently the not-allowed cursor is applied directly to the host so I needed something at that level.  Could also have used an event instead of a callback to sync the parent state.  Also wondering whether I should just do the `requireParent` on initialize, but I worry with Angular content projection it might be possible for a tab to be rendered detached.  Not sure if there's a simple way to ensure the parent reference is kept up to date if the tab is attached after initialization.

In Forge 2.0, the TabBar did not have a disabled property, and developers would have needed to set disabled on each tab individually, and they could have used something like `parentDisabledCondition || individualTabCondition`.  So I did consider just making it so that if the parent is not disabled, it wouldn't initially override the children, so if you wanted to handle both you'd need to do the same as before, and just not use them together, but I felt like even if documented, that would lead to confusion and questions because developers wouldn't find that intuitive. 

## Additional information
[Teams Discussion](https://teams.microsoft.com/l/message/19:1gCw-wLjrS56X_iAY0mpHpk4PyWOi08JiMM7Kc1xvho1@thread.tacv2/1729105258525?tenantId=7cc5f0f9-ee5b-4106-a62d-1b9f7be46118&groupId=2bb1b8ea-cd12-4df5-a4ab-0c895bfcda5c&parentMessageId=1729105258525&teamName=Forge&channelName=General&createdTime=1729105258525)

### DIFF
--- a/src/lib/tabs/tab-bar/tab-bar-core.ts
+++ b/src/lib/tabs/tab-bar/tab-bar-core.ts
@@ -180,11 +180,11 @@ export class TabBarCore implements ITabBarCore {
   private _syncTabState(): void {
     this._tabs.forEach((tab, index) => {
       tab.selected = index === this._activeTab;
-      tab.disabled = this._disabled;
       tab.vertical = this._vertical;
       tab.stacked = this._stacked;
       tab.secondary = this._secondary;
       tab.inverted = this._inverted;
+      tab.syncTabDisabledState();
     });
   }
 
@@ -263,7 +263,7 @@ export class TabBarCore implements ITabBarCore {
     value = Boolean(value);
     if (this._disabled !== value) {
       this._disabled = value;
-      this._tabs.forEach(tab => (tab.disabled = this._disabled));
+      this._tabs.forEach(tab => tab.syncTabDisabledState());
       this._adapter.toggleHostAttribute(TAB_BAR_CONSTANTS.attributes.DISABLED, this._disabled);
     }
   }
--- a/src/lib/tabs/tab/tab-adapter.ts
+++ b/src/lib/tabs/tab/tab-adapter.ts
@@ -1,9 +1,10 @@
-import { getShadowElement, requireParent, toggleAttribute } from '@tylertech/forge-core';
+import { getShadowElement, requireParent, toggleAttribute, toggleClass } from '@tylertech/forge-core';
 import { IStateLayerComponent, STATE_LAYER_CONSTANTS } from '../../state-layer';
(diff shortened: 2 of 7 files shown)


MikeMatusz:
After discussion, going with the simpler approach of just not supporting a mixed usage of tab/tab-bar.  Going to create a new PR momentarily rather than update this one.

PR #720, opened 2024-10-22
fix(field): use error color for support text when field is invalid

samrichardsontylertech:
## PR Checklist

## Describe the new behavior?
Fixes a regression where support text wasn't changing to the error color when a field is invalid.

### DIFF
--- a/src/lib/field/field.scss
+++ b/src/lib/field/field.scss
@@ -173,6 +173,10 @@ $variants: (
   .forge-field {
     @include core.theme('error');
     @include core.invalid-label-color;
+
+    .support-text {
+      @include core.invalid-support-text-color;
+    }
   }
 }
 
--- a/src/lib/field/_core.theme.scss
+++ b/src/lib/field/_core.theme.scss
@@ -161,3 +161,7 @@ $elevations: (
 @mixin invalid-label-color {
   @include override(label-color, theme.variable(error), value);
 }
+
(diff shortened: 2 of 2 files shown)


Ross-Blakeney (approved):

DRiFTy17 (approved):

src/lib/field/field.scss
MikeMatusz, line 1: Is it intentional that the label color is different between `invalid` and `theme="error"`?  And if we're coloring support text for the invalid state, should it be colored for all semantic themes?

MikeMatusz: I definitely agree this makes sense for an invalid field, since support-text is typically used to display the error.  Semantic themes are a separate issue, and more of a design question.  Shouldn't hold up this change, just trying to determine what changes were intended.

I still think it's odd that the label color differs between `error` and `invalid`, though. Have we verified contrast with filled variants in both dark and light themes?  I think `invalid` label contrast is only 5.17:1 for filled fields in dark theme.  That's pre-existing, though, so could certainly be a separate PR to address that if needed.

samrichardsontylertech: Yeah, I'm fine with whatever the consensus is on the design. Not sure about the contrast but I'm not surprised that it might be an issue. Our theme colors are limited when it comes to all the states this supports. What I really want to do is remove the `inset` label position altogether. That would solve half our problems immediately.

PR #722, opened 2024-10-23
fix: updated card and scaffold to allow for overflowing content by default

DRiFTy17:
## PR Checklist

- Does this PR introduce a breaking change? Potentially, but this is a more sane default for the majority of use cases.

## Describe the new behavior?
Updates the card and scaffold internal `overflow` styles to default to `initial` (allow overflowing content) instead of `hidden`.

## Additional information
This change _could_ potentially cause inadvertent overflow in specific layouts that rely on it, but in initial testing this would likely be pretty rare, and it is much more likely (and predictable) that developers want overflow by default and instead manually set `hidden` if desired via the component specific tokens. This is consistent with how default CSS works.

### DIFF
--- a/src/lib/core/styles/tokens/card/_tokens.scss
+++ b/src/lib/core/styles/tokens/card/_tokens.scss
@@ -16,7 +16,7 @@ $tokens: (
   elevation: utils.module-val(card, elevation, none),
   padding: utils.module-val(card, padding, spacing.variable(medium)),
   shape: utils.module-val(card, shape, shape.variable(medium)),
-  overflow: utils.module-val(card, overflow, hidden),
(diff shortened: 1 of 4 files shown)


samrichardsontylertech (approved):

DRiFTy17:
Moving back to draft until we can vet this change a bit more before introducing it. So far, no side effects were found, but it's not necessarily the crux of the issue with clipped focus indicators. We'll instead provide guidance in the docs for how to handle those scenarios, and consider introducing this change in a future release.

DRiFTy17:
Closing in favor of https://github.com/tyler-technologies-oss/forge/pull/736

PR #754, opened 2024-11-19
fix(field): constrain input to grid area to prevent icons from being pushed out of view

### LINKED ISSUE #749: [chip-field] Popover icon disappears when the field is small and a selection is made
**Describe the bug:**
The popover icon at the end of the chip field disappears when at least one item is selected and the field width is small.

**To Reproduce:**
1. Enter a value in a chip field with a popover icon
2. Make your window smaller, or reduce the width of the field, or zoom your window in significantly.

See: https://forge.tylerdev.io/main/?path=/story/components-chip-field--demo&args=popoverIcon:!true

**Expected behavior:**
The icon should be visible.

**Screenshots:**
image

 - Forge version: 3.3.5
 - Browser: All
 - Platform: All
 - OS: All

MikeMatusz:
Fixes #749 

## PR Checklist

- Tests for the changes have been added/updated: n/a (visual)
- Docs have been added/updated: n/a

## Describe the new behavior?
Because `min-inline-size` defaults to auto, the default size of the input can grow beyond the bounds of the CSS grid layout area assigned to it, pushing icon slots outside the bounds of the container.  By explicitly setting it to zero, the input will be sized based on available space as the assigned `1fr` intends.  

## Additional information
Long labels can still push the icons, though the behavior differs in Firefox vs Chrome.  This is a more complex problem with usability/accessibility concerns, and I have not attempted to resolve this.

### DIFF
--- a/src/lib/field/_core.scss
+++ b/src/lib/field/_core.scss
@@ -127,6 +127,7 @@
 
   block-size: 100%;
   inline-size: 100%;
+  min-inline-size: 0;
 
   color: theme.variable(text-high);
 }


MikeMatusz:
@DRiFTy17 This appears to be another side effect of #730.  Would fixing the width of the overall container, instead of the `input` container specifically, have also worked? Or can you provide an example or elaborate on the the original scenario for why that wouldn't suffice?

conner-fallone (approved):

samrichardsontylertech (approved):

PR #756, opened 2024-11-20
feat: enhance skip-link defaults

Ross-Blakeney:
## PR Checklist

## Describe the new behavior?

Summary: enhance the skip-link defaults to improve ease of use

- Auto add a `tabindex="-1"` to the target element if a tabindex does not already exist
- If a `target` is not supplied, automatically find and use the first main element as the target. This will use that element's ID, or add an ID if that element does not have one.

## Additional information

Looking for feedback before i try to take this any further. Maybe this approach, which includes setting attributes on external elements, is against the Forge ethos.

This assumes that there will only be a single main element present at a time. Maybe that's not a safe enough assumption.

**WIP - not in a state that I expect to merge.**

### DIFF
--- a/src/dev/pages/skip-link/skip-link.ejs
+++ b/src/dev/pages/skip-link/skip-link.ejs
@@ -2,8 +2,14 @@
   <section>
     <h3>Default</h3>
     <div tabindex="0">Tab past this</div>
-    <forge-skip-link id="skip-link" target="main"></forge-skip-link>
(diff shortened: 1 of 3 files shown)


Ross-Blakeney:
This clearly needs work to actually function. Just feeling the waters.
For example, that approach to initializing a target if it is not provided would not currently work in a SPA where the main element would be getting destroyed on every route event.

Ross-Blakeney:
This is too intrusive for Forge functionality. Closing.

PR #763, opened 2024-11-21
fix(field): fix input size/animation when floating label with inset variant

### LINKED ISSUE #752: [field] icons visibly shift when end icon added on blur
**Describe the bug:**
We have a pattern where we display an `end` slot icon if a field is `invalid`, but following best practices only display the validation state after the field becomes touched, which happens on blur.  There is a noticeable shift in the icons, possibly due to the float label animation fixing the width of the input.  This could potentially happen on any field, but is noticeable on date (and probably time) because if `showMaskFormat` is true, or you type an invalid date and tab off, blur also clears the input.  If Date is a required field (which it commonly is for us) this occurs frequently.

A way to reproduce it on any generic field is to have an end icon and `showClear`, and then type some text and backspace it out.  The icons also shift in this scenario.

**To Reproduce:**
1. Go to the [Forge Date Picker example](https://forge.tylerdev.io/main/?path=/story/components-date-picker--demo&args=showMaskFormat:!true), and ensure `showMaskFormat` is toggled on.
2. Select the `<input>` element in dev tools, and run the following snippet in the console:
```js
$0.addEventListener('blur', () => {
    const icon = document.createElement('forge-icon');
    icon.name = 'insert_invitation';
    icon.slot = 'end';
    document.querySelector('forge-text-field').appendChild(icon);
});
``` 
3. Focus and blur the date picker field, and note the icons shift.

**Expected behavior:**
The icon should appear where it belongs, and not shift afterwards.

**Screenshots:**
date-picker-blur

 - Forge version: 3.x
 - Browser: All
 - Platform: All
 - OS: All

DRiFTy17:
## PR Checklist

## Describe the new behavior?
This change removes the internal `width` being set while the inset floating label animation is running, but instead switches the `<input>` styles to use a simpler `padding-block-start` transition. This ensures that the input container doesn't use `position: absolute` any longer, but does come with the caveat of animating padding which is less performant.

## Additional information
Fixes #752

### DIFF
--- a/src/dev/pages/text-field/text-field.ts
+++ b/src/dev/pages/text-field/text-field.ts
@@ -42,8 +42,19 @@ const optLabelAlignment = document.getElementById('opt-label-alignment') as ISel
 const optSupportTextInset = document.getElementById('opt-support-text-inset') as ISwitchComponent;
 
 optLabel.addEventListener('input', () => {
+  const hasValue = optLabel.value.length > 0;
   textFields.forEach(textField => {
-    textField.querySelector(':is(label, span, forge-label)').textContent = optLabel.value;
+    const labelEl = textField.querySelector(':is(label, span, forge-label)');
+    if (hasValue) {
+      if (!labelEl) {
+        const label = document.createElement('label');
+        label.textContent = optLabel.value;
+        textField.prepend(label);
+      }
+      textField.querySelector(':is(label, span, forge-label)').textContent = optLabel.value;
+    } else {
+      labelEl?.remove();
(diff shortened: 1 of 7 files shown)


MikeMatusz (approved):
Verified in both the test app and dropping the dist folder into node_modules of a large hybrid Angular/AngularJS app using PSDF.  This definitely resolves the linked issue, and I haven't seen any other problems.  Even resetting a large form with/without data to trigger float animations en masse, I'm not seeing any observable lag.

PR #766, opened 2024-11-26
feat(tabs): deprecate `secondary` prop/attr and update default design to use full width tab indicator

DRiFTy17:
## PR Checklist

- Does this PR introduce a breaking change? N (`secondary` APIs exist, but have no effect)

## Describe the new behavior?
After discussion with the Forge design governance committee, it was determined that we should combine the "primary" (default) and "secondary" variants into a single design. This is purely related to the indicator, and the new (and only) default design can be seen below.

The `secondary` property and attribute still exist, as well as any related design tokens, but they have no effect on the design. I've marked them as deprecated. It was done this way to avoid breaking any existing code. We'll plan to remove them in a future version.

Indicator design updates:
- Expand indicator to be full width of tab (from "secondary" design)
- Use "primary" indicator height (`3px` by default)
- Use "primary" indicator shape (rounded corners)

These same changes will also been apply to the vertical and inverted variants.

Example of changes:

image

### DIFF
--- a/src/dev/pages/tabs/tabs.ejs
+++ b/src/dev/pages/tabs/tabs.ejs
@@ -48,7 +48,7 @@
 </div>
 
 <div class="tabs-demo">
-  <h2 class="forge-typography--heading2">Primary tabs w/icons</h2>
+  <h2 class="forge-typography--heading2">Stacked Tabs w/icons</h2>
   <forge-tab-bar stacked active-tab="0" style="max-width: 512px;">
     <forge-tab>
       <forge-icon slot="leading" name="videocam"></forge-icon>
--- a/src/dev/pages/tabs/tabs.ejs
+++ b/src/dev/pages/tabs/tabs.ejs
@@ -65,14 +65,4 @@
   </forge-tab-bar>
 </div>
 
-<div class="tabs-demo">
-  <h2 class="forge-typography--heading2">Secondary tabs</h2>
-  <forge-tab-bar secondary active-tab="0">
-    <forge-tab>Overview</forge-tab>
-    <forge-tab>Specs</forge-tab>
-    <forge-tab>Guidelines</forge-tab>
-    <forge-tab>Accessibility</forge-tab>
-  </forge-tab-bar>
-</div>
-
 <script type="module" src="tabs.ts"></script>
--- a/src/dev/pages/tabs/tabs.html
+++ b/src/dev/pages/tabs/tabs.html
@@ -5,7 +5,6 @@ include('./src/partials/page.ejs', {
     includePath: './pages/tabs/tabs.ejs',
     options: [
       { type: 'switch', label: 'Vertical', id: 'opt-vertical' },
-      { type: 'switch', label: 'Secondary', id: 'opt-secondary' },
       { type: 'switch', label: 'Clustered', id: 'opt-clustered' },
       { type: 'switch', label: 'Stacked', id: 'opt-stacked' },
       { type: 'switch', label: 'Disabled', id: 'opt-disabled' },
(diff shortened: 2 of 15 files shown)


Ross-Blakeney (approved):

samrichardsontylertech:
Looks good 👍

Maybe we should add a little bit of inline padding so when the indicator is against an edge the radius has breathing room?

tylermcgowan:
> We had briefly discussed how the rounded "shape" of the indicator at full width now may conflict with our expectations when placed against a hard edge. Is this still a valid concern?
> 
> The shape can be controlled via design token to square the indicator with a `0` radius, but if this is a desired/common scenario we _need_ to account for, we could introduce an easy-to-use API for controlling that, but I'll wait for feedback before determining if that's the case.

I don't feel the radius is too much concern for preserving a hard left edge. It's still there visually, and there's also the `tab-bar-divider` helping keep that edge when other tabs are selected.

If we wanted to keep a secondary variant/option, it could be with the text left aligned. Could we also bump the font-weight to 600 (semibold) for the active tab?

CleanShot 2024-11-26 at 10 29 02@2x

DRiFTy17:
> I don't feel the radius is too much concern for preserving a hard left edge. It's still there visually, and there's also the `tab-bar-divider` helping keep that edge when other tabs are selected.

Sounds good. We can always come back and adjust it in the future if it becomes a problem.

> If we wanted to keep a secondary variant/option, it could be with the text left aligned. 

Let's revisit this in a separate change if there's a need since it's a new variant that would need some guidance on when to use it.

> Could we also bump the font-weight to 600 (semibold) for the active tab?

While I do agree, the problem with adjusting the `font-weight` is that it can very slightly affect the layout when the tabs are clustered... This happens because the size required for the label is dependent on its content, which when it grows due to the font-weight changing would move the other tabs over slightly... We could potentially use `text-shadow` instead, but the text won't be crisp... @samrichardsontylertech any other ideas on this one?

> And for the icon variant, could we increase the column-gap to 8px?

Yes. I'll make that change.

samrichardsontylertech:
> While I do agree, the problem with adjusting the font-weight is that it can very slightly affect the layout when the tabs are clustered... This happens because the size required for the label is dependent on its content, which when it grows due to the font-weight changing would move the other tabs over slightly... We could potentially use text-shadow instead, but the text won't be crisp... @samrichardsontylertech any other ideas on this one?

@DRiFTy17  Let's not use `text-shadow`, that's going to look pretty bad. Instead we could reserve the space with a transparent pseudo element with the same content permanently bolded. It's kind of hacky but probably the best solution if we're determined to do this.

samrichardsontylertech:
Or we could switch to a uniwidth font 🤷‍♂️
[https://uxdesign.cc/uniwidth-typefaces-for-interface-design-b6e8078dc0f7](https://uxdesign.cc/uniwidth-typefaces-for-interface-design-b6e8078dc0f7)

DRiFTy17:
> Or we could switch to a uniwidth font 🤷‍♂️ https://uxdesign.cc/uniwidth-typefaces-for-interface-design-b6e8078dc0f7

Yeah we discussed switching to Roboto Flex a while back. Maybe we push off this change until we're ready to introduce support for a new font. This would be a good item for v4, but we could probably introduce it sooner than that.

tylermcgowan:
> While I do agree, the problem with adjusting the font-weight is that it can very slightly affect the layout when the tabs are clustered...

Ahh yeah, good call @DRiFTy17, I overlooked that bit. I'm with Sam on this as well... text-shadow won't be great, so we can revisit later.

> Yeah we discussed switching to Roboto Flex a while back. Maybe we push off this change until we're ready to introduce support for a new font. This would be a good item for v4, but we could probably introduce it sooner than that.

 👆🏻 I'm ready to help with this whenever you are! I had done the revised type scale with Roboto Flex, so it should be nearly good to go. Let me know if/when you're ready to start on it and I'll make sure to actually get you the CSS for it.

samrichardsontylertech (approved):

Issue #774, opened 2024-12-30
[expansion-panel] Add a "target-button" attribute to associate the expansion panel with a button


eliganemtyler:
**Describe the feature:**
A `<forge-expansion-panel>` could optionally be associated with a button element that isn't in its `header` slot via a new `target-button` attribute with a value of the button's `id`, similarly to how a `forge-tooltip` can be associated with an anchor element via the `anchor` attribute. Setting the `target-button` of a `<forge-expansion-panel>` should allow the component to automatically associate the button's `aria-controls` to the element in the `default` slot of the panel, handle toggling the panel via the button the way it does now if the button is in the `header` slot, and manage the `aria-expanded` state of the button.

**Describe any alternatives you've considered:**
n/a
**Additional context:**
n/a

PR #775, opened 2024-12-30
Feat/expansion panel target button

eliganemtyler:
## PR Checklist

## Describe the new behavior?
This PR adds a `target-button` attribute and property to the expansion panel that matches the description in the linked issue. The `target-button` value should match the id of an element in the document intended to toggle the panel. If `target-button` isn't used (or the element with the matching `id` can't be found), the `<button>` or element with `button` role in the `header` slot is the "target button". The target button toggles the `open` state of the component, and has `aria-controls` set by and `aria-expanded` managed by the component. The element in the default slot is assigned a unique id if it doesn't already have one, to be linked to by `aria-controls`.

## Additional information
For simplicity's sake, the new feature doesn't support multiple target buttons. Because a panel using `target-button` is setting the sole target button to one outside the component, the `header` part of the `forge-expansion-panel` using `target-button` attribute gets `cursor: auto` instead of `cursor: pointer`. Configurations with multiple buttons toggling the expansion panel would be the responsibility of the developer to manage.


eliganemtyler:
It looks like I don't have permission to link the issue to this PR, but it's this one: https://github.com/tyler-technologies-oss/forge/issues/774

Let me know what you think!

src/lib/expansion-panel/expansion-panel.ts
```diff
@@ -11,6 +11,7 @@ export interface IExpansionPanelComponent extends IBaseComponent {
   open: boolean;
   orientation: ExpansionPanelOrientation;
   animationType: ExpansionPanelAnimationType;
+  targetButton: string;
```
DRiFTy17, line 14: Let's rename the `targetButton` property and `target-button` attribute to something like `trigger`? Open to other suggestions.

src/lib/expansion-panel/expansion-panel-adapter.ts
```diff
@@ -7,16 +7,24 @@ import { EXPANSION_PANEL_CONSTANTS } from './expansion-panel-constants';
...
+  detachAria(): void;
 }
 
 export class ExpansionPanelAdapter extends BaseAdapter<IExpansionPanelComponent> implements IExpansionPanelAdapter {
   private _headerElement: HTMLElement;
   private _contentElement: HTMLElement;
   private _innerElement: HTMLElement;
+  private _headerSlotElement: HTMLSlotElement;
+  private _defaultSlotElement: HTMLSlotElement;
+  private _targetButton: HTMLElement | null;
```
DRiFTy17, line 27: We should not hold on to the element reference to avoid memory leaks due to the disconnected nature of this change. The problem is that there is no way for us to know if the associated trigger element is removed from the DOM without using a `MutationObserver`.

What we should do instead is get rid of this private property and dynamically attempt to locate the trigger element when attaching the listeners via either initialization of the component, or when the trigger IDREF changes.

In terms of a11y attributes, we should also dynamically locate the trigger element each time the open state is toggled before setting those attributes.

src/lib/expansion-panel/expansion-panel-adapter.ts
```diff
@@ -39,6 +49,12 @@ export class ExpansionPanelAdapter extends BaseAdapter<IExpansionPanelComponent>
     this._headerElement.addEventListener(type, listener);
   }
 
+  public addTargetButtonListener(type: keyof HTMLElementEventMap, listener: EventListener): void {
+    if (this._targetButton) {
+      this._targetButton.addEventListener(type, listener);
+    }
```
DRiFTy17, line 55: Let's use optional chaining here instead as it's more modern:
```javascript
this._targetButton?.addEventListener(type, listener);
```

src/lib/expansion-panel/expansion-panel-adapter.ts
```diff
@@ -70,4 +86,44 @@ export class ExpansionPanelAdapter extends BaseAdapter<IExpansionPanelComponent>
   public animationEnd(): void {
     this._innerElement.style.removeProperty('overflow');
   }
+
+  public tryLocateTargetButton(id: string): void {
+    this._targetButton = this._tryFindTargetButton(id);
```
DRiFTy17, line 91: We won't need this method any more because we're not storing the element reference and the core file will not need to call this.

src/lib/expansion-panel/expansion-panel-adapter.ts
```diff
@@ -70,4 +86,44 @@ export class ExpansionPanelAdapter extends BaseAdapter<IExpansionPanelComponent>
...
+  }
+
+  public updateAriaExpanded(open: boolean): void {
+    if (this._targetButton) {
+      if (open) {
+        this._targetButton.setAttribute('aria-expanded', 'true');
+      } else {
+        this._targetButton.setAttribute('aria-expanded', 'false');
+      }
+    }
```
DRiFTy17, line 110: We could use optional chaining here as well, or you could use the `toggleAttribute()` helper from `@tylertech/forge-core` if you prefer. It allows for toggling and setting a value in a one liner.

src/lib/expansion-panel/expansion-panel-adapter.ts
```diff
@@ -70,4 +86,44 @@ export class ExpansionPanelAdapter extends BaseAdapter<IExpansionPanelComponent>
...
+  }
+
+  public detachAria(): void {
+    if (this._targetButton) {
+      this._targetButton.removeAttribute('aria-controls');
+      this._targetButton.removeAttribute('aria-expanded');
+    }
```
DRiFTy17, line 117: More optional chaining

src/lib/expansion-panel/expansion-panel-core.ts
```diff
@@ -110,4 +122,14 @@ export class ExpansionPanelCore implements IExpansionPanelCore {
...
     }
   }
+
+  public get targetButton(): string {
+    return this._targetButton;
+  }
+  public set targetButton(value: string) {
+    if (this._targetButton !== value) {
+      this._targetButton = value;
+      this._adapter.setHostAttribute(EXPANSION_PANEL_CONSTANTS.attributes.TARGET_BUTTON, this._targetButton);
```
DRiFTy17, line 132: We will need additional logic here to remove any listeners from a previously attached element in case the trigger button changes dynamically and one is already set at that time. We should write a test care for this scenario as well.

src/lib/expansion-panel/expansion-panel-core.ts
```diff
@@ -22,7 +24,16 @@ export class ExpansionPanelCore implements IExpansionPanelCore {
...
+    this._adapter.tryLocateTargetButton(this._targetButton);
+    this._adapter.addTargetButtonListener('click', this._clickListener);
+    this._adapter.addTargetButtonListener('keydown', this._keydownListener);
     this._adapter.setAnimationCompleteListener(this._animationCompleteListener);
+    this._adapter.initializeAccessibility();
+    this._adapter.updateAriaExpanded(this._open);
+  }
+
+  public destroy(): void {
+    this._adapter.detachAria();
```
DRiFTy17, line 36: We should attempt to remove the listeners from the detached trigger element when the expansion panel is destroyed.

src/lib/expansion-panel/expansion-panel-core.ts
```diff
@@ -22,7 +24,16 @@ export class ExpansionPanelCore implements IExpansionPanelCore {
   public initialize(): void {
     this._adapter.addHeaderListener('click', this._clickListener);
     this._adapter.addHeaderListener('keydown', this._keydownListener);
+    this._adapter.tryLocateTargetButton(this._targetButton);
+    this._adapter.addTargetButtonListener('click', this._clickListener);
+    this._adapter.addTargetButtonListener('keydown', this._keydownListener);
```
DRiFTy17, line 29: We may need to need to delay this by a `requestAnimationFrame()` because in frameworks like Angular that use an incremental rendering approach, or if DOM is added between frames in general, the trigger element may not be available in the DOM yet.

This one has come up a lot in various components in the past and we've learned that testing in Angular uncovers a lot of timing issues due to incremental rendering.

DRiFTy17 (changes requested):
The changes look great overall. This should really assist developers with convenience, and we'll want to make sure we let everyone know when it's available.

I left some comments below of things that jump out at me, let me know if something doesn't make sense and feel free to push back on anything! I'll do a more in depth review/testing of usage when the next set of changes are pushed up.

src/lib/expansion-panel/expansion-panel-core.ts
eliganemtyler: This part of the "should update old and new triggers on trigger change" test will test the event listeners being removed:

```
trigger1.click();
expect(expansionPanel.open).to.be.true;

expansionPanel.trigger = 'button-id2';
expect(trigger1.getAttribute('aria-controls')).to.be.null;
expect(trigger1.getAttribute('aria-expanded')).to.be.null;
trigger1.click();
expect(expansionPanel.open).to.be.true; //expansion panel stays open when old trigger is clicked
```

src/lib/expansion-panel/expansion-panel-core.ts
eliganemtyler: I think I've got `requestAnimationFrame()` in the right place, but I'm not completely sure. It's wrapping around the `_syncTrigger()` call during `initialize()`. I wasn't sure if there was a way to write a unit test for this so that I could really see if this would prevent the incremental rendering issue you're describing.

eliganemtyler:
My recent commits hopefully address all the points you raised @DRiFTy17.

In addition to better handling cases where the `trigger` id changes and points to another element, I also handle cases where the slotted content changes with a `slotchange` listener. I saw that in some components a `slotchange` listener gets manually removed on component destroy (e.g. button-area) and in some it doesn't (e.g. app-bar). My assumption is that it's fine the latter way because the listeners would get cleaned up by js once the component is destroyed but let me know if I'm wrong.

Also, I wanted to make sure `document` reference in `_tryLocateTriggerElement()` was okay. I was originally using `this._component.getRootNode()` there, but by the time it was called during the disconnect hook it didn't work. I saw precedent for using `document` in the skip-link component.

Thanks!

src/lib/expansion-panel/expansion-panel-constants.ts
```diff
@@ -1,3 +1,4 @@
+import { triggerAsyncId } from 'async_hooks';
```
DRiFTy17, line 1: This looks like an accidental import that we can remove.

src/lib/expansion-panel/expansion-panel-core.ts
```diff
@@ -5,24 +5,50 @@ export interface IExpansionPanelCore {
...
 
   public initialize(): void {
     this._adapter.addHeaderListener('click', this._clickListener);
     this._adapter.addHeaderListener('keydown', this._keydownListener);
     this._adapter.setAnimationCompleteListener(this._animationCompleteListener);
+    this._adapter.addContentSlotListener(this._slotListener);
+    this._adapter.setContentId();
+    requestAnimationFrame(() => {
+      this._syncTrigger();
+    });
```
DRiFTy17, line 33: You can import the `frame()` helper and use it like this:
```typescript
await frame();
this._syncTrigger();
```
It's just a `Promise` wrapper around `requestAnimationFrame()` for better ergonomics. You'll just need to make the `initialize()` method async which is fine.

When you write tests for this, you just need to await the frame before any assertions to ensure the rAF has time to finish. There should be a lot of examples in existing tests that do this, or use `elementUpdated()` as well.

src/lib/expansion-panel/expansion-panel-core.ts
```diff
@@ -110,4 +137,16 @@ export class ExpansionPanelCore implements IExpansionPanelCore {
...
+
+  public get trigger(): string {
+    return this._trigger;
+  }
+  public set trigger(value: string) {
+    if (this._trigger !== value) {
+      this._adapter.detachTriggerAria();
+      this._trigger = value;
+      this._syncTrigger();
+      this._adapter.setHostAttribute(EXPANSION_PANEL_CONSTANTS.attributes.TRIGGER, this._trigger);
```
DRiFTy17, line 149: We're starting to get away from reflecting properties to attributes actually. Let's go ahead and remove this call to `setHostAttribute()` so that we don't have to remove it later when we do a sweep across the library.

src/lib/expansion-panel/expansion-panel.test.ts
```diff
@@ -419,6 +422,153 @@ describe('Expansion Panel', () => {
...
+        <div>
+          <button id="button-id"></button>
+          <forge-expansion-panel trigger="button-id"></forge-expansion-panel>
+        </div>
+      `);
+    });
+
+    it('should not error if trigger not found', async () => {
+      await fixture<IExpansionPanelComponent>(html`<forge-expansion-panel trigger="button-id"></forge-expansion-panel>`);
+    });
```
DRiFTy17, line 569: Tests without assertions are typically bad practice (even though we know that exceptions can cause a test to fail) but most of the time these result in silent errors and false positives.

src/stories/components/expansion-panel/ExpansionPanel.mdx
```diff
@@ -20,36 +20,45 @@ It's common to compose an expansion panel with a card component to provide a mor
 
 > **Tip:** You can provide a `<forge-open-icon>` to show the open state of the expansion panel, and the icon will be rotated when the panel is opened/closed.
 
+## Target Button
+
+A button in the header slot of the expansion panel will automatically control toggling the `open` state of the panel.
```
DRiFTy17, line 25: May be worth adding a note about ensuring to control the existence of the trigger button and the panel in the DOM at the same time to avoid any hard to debug issues when it comes to async/incremental rendering.

These types of scenarios are why we may want to consider adding a `triggerElement` property to allow devs to set the instance directly in cases where timing can be an issue.

Eventually we can invert this process by using the Invoker Commands API with a custom command so that the button will instead notify the panel rather than the other way around which will simplify things a lot.

src/lib/expansion-panel/expansion-panel-adapter.ts
```diff
@@ -35,10 +63,25 @@ export class ExpansionPanelAdapter extends BaseAdapter<IExpansionPanelComponent>
...
 
+  public addContentSlotListener(listener: EventListener): void {
+    this._defaultSlotElement.addEventListener('slotchange', listener);
+  }
+
   public addHeaderListener(type: keyof HTMLElementEventMap, listener: EventListener): void {
     this._headerElement.addEventListener(type, listener);
   }
 
+  public removeTriggerListeners(reset = false): void {
```
DRiFTy17, line 74: Just a small code style request when it comes to parameters of methods. Using the "config object" pattern can typically lead to more readable code. Something like this:
```typescript
public removeTriggerListeners({ reset } = { reset: false }): void {
```

Then when calling it things are bit more clear:
```typescript
this._adapter.removeTriggerListeners({ reset: true });
```

src/lib/expansion-panel/expansion-panel.ts
```diff
@@ -98,6 +107,9 @@ export class ExpansionPanelComponent extends BaseComponent implements IExpansion
   @coreProperty()
   public declare animationType: ExpansionPanelAnimationType;
 
+  @coreProperty()
+  public declare trigger: string;
```
DRiFTy17, line 111: See other comment about potentially adding a `triggerElement` property so that devs can manually set the trigger element instance to avoid race condition issues. We have run into this in Angular for other similar properties and this is usually preferred due to the async nature of incremental rendering (which is just the process of adding an element, waiting a frame, then adding another element which causes the component initialization to need to wait for discovery). Using direct element references fixes this problem and gives a nice alternative usage.

Once again the Invoker Commands API will eventually supersede this pattern completely by inverting the communication.

DRiFTy17 (commented):
Sorry for the delay, but here's another quick review. Things are looking great so far!

eliganemtyler:
Hey @DRiFTy17, I've incorporated your feedback into these latest changes (everything from "remove unused import" on).

I added the `triggerElement` property. I used the `anchor` and `anchorElement` pattern in tooltip as a reference, but it was a little different since tooltip sets and maintains a reference to the anchor element, whereas the expansion panel always looks for the trigger element dynamically by its `id` each time it needs it.

I considered the case where both `trigger` (id) and `triggerElement` (element ref) are set (which would be an incorrect use). I decided to have the component simply throw an error in that case, instead of worrying about which should take precedent.

src/lib/expansion-panel/expansion-panel-adapter.ts
```diff
@@ -1,22 +1,35 @@
...
+  updateAriaControls(): void;
+  updateAriaExpanded(open: boolean): void;
+  detachTriggerAria(): void;
 }
 
 export class ExpansionPanelAdapter extends BaseAdapter<IExpansionPanelComponent> implements IExpansionPanelAdapter {
   private _headerElement: HTMLElement;
   private _contentElement: HTMLElement;
   private _innerElement: HTMLElement;
+  private _headerSlotElement: HTMLSlotElement;
```
DRiFTy17, line 29: It looks like this is unused, let's go ahead and remove it.

src/lib/expansion-panel/expansion-panel-core.ts
```diff
@@ -1,28 +1,55 @@
...
+    this._adapter.detachTriggerAria();
+    this._adapter.removeTriggerListeners();
+  }
+
+  private _handleContentSlotChange(): void {
+    this._adapter.setContentId();
+    this._syncTrigger();
+  }
+
+  private _syncTrigger(): void {
```
DRiFTy17, line 47: We may want to handle the scenario where the trigger element is detached from the DOM dynamically and release our internal reference to it each time this method runs by checking if the trigger element ref is not connected.

src/lib/expansion-panel/expansion-panel-core.ts
```diff
@@ -110,4 +138,36 @@ export class ExpansionPanelCore implements IExpansionPanelCore {
...
+    if (this._adapter.triggerElementRef !== el) {
+      this._adapter.detachTriggerAria();
+      this._adapter.setTriggerElement(el);
+      this._syncTrigger();
+    }
+  }
+
+  private _throwTriggerError(): void {
+    throw new Error('trigger and triggerElement cannot both be set.');
+  }
```
DRiFTy17, line 172: I honestly wouldn't even bother throwing an exception in this scenario... I think if someone is choosing the more explicit `triggerElement` property to control it then it should take precedence IMO. It should be very rare if someone uses these two together, and it's probably ok to assume the element reference is of higher priority. I could be convinced otherwise though if there's good reason to throw.

src/lib/expansion-panel/expansion-panel-adapter.ts
```diff
@@ -70,4 +120,40 @@ export class ExpansionPanelAdapter extends BaseAdapter<IExpansionPanelComponent>
...
+  }
+
+  public detachTriggerAria(): void {
+    this._triggerElement?.removeAttribute('aria-controls');
+    this._triggerElement?.removeAttribute('aria-expanded');
+  }
+
+  private _tryLocateTriggerElement(id: string): HTMLElement | null {
+    if (id) {
+      const triggerEl = document.getElementById(id);
```
DRiFTy17, line 153: Other components may have this same issue, but we should instead get the root node to ensure it is scoped to the specific DOM tree given the potential for other shadow root boundaries (if we were to use the expansion panel within the shadow of another element then it wouldn't work):
```javascript
const triggerEl = this._component.getRootNode().getElementById(id);
```

src/lib/expansion-panel/expansion-panel-adapter.ts
```diff
@@ -27,6 +40,28 @@ export class ExpansionPanelAdapter extends BaseAdapter<IExpansionPanelComponent>
...
+  private get _triggerElement(): HTMLElement | null {
+    return this._triggerElementRef ?? this._tryLocateTriggerElement(this._component.trigger);
+  }
+
+  public get triggerElementRef(): HTMLElement | null {
+    return this._triggerElementRef;
+  }
+
+  private get _slottedContentElement(): Element | undefined {
+    return this._defaultSlotElement.assignedElements()[0];
```
DRiFTy17, line 56: We should use `assignedElements({ flatten: true })` here instead to ensure that it handles composed slot elements (slots acting as passthrough slots on potential parent elements).

src/lib/expansion-panel/expansion-panel-adapter.ts
```diff
@@ -27,6 +40,28 @@ export class ExpansionPanelAdapter extends BaseAdapter<IExpansionPanelComponent>
...
+  }
+
+  private set _slottedContentId(id: string) {
+    this._slottedContentElement?.setAttribute('id', id);
```
DRiFTy17, line 60: I think we only want to do this if the slotted element doesn't already have an `id`?

src/lib/expansion-panel/expansion-panel-adapter.ts
```diff
@@ -70,4 +120,40 @@ export class ExpansionPanelAdapter extends BaseAdapter<IExpansionPanelComponent>
   public animationEnd(): void {
     this._innerElement.style.removeProperty('overflow');
   }
+
+  public setTriggerElement(el: HTMLElement | null): void {
+    this._triggerElementRef = el;
+  }
+
+  public setContentId(): void {
+    if (!this._slottedContentId) {
```
DRiFTy17, line 129: We should probably add a check to use the existing `id` on the slotted element if it already has one? Currently this overwrites the developer-provided `id` which isn't what we want.

DRiFTy17 (changes requested):
@eliganemtyler It's coming along nicely, this iteration of changes look to be much better IMO. I pulled this down and gave it some thorough testing and made a few notes below. We're getting there!

src/lib/expansion-panel/expansion-panel-adapter.ts
eliganemtyler: `if (!this._slottedContentId)` in `setContentId()` checks for this before we call this setter.

src/lib/expansion-panel/expansion-panel-adapter.ts
eliganemtyler: See above comment. I also have a test for this: "should not overwrite existing id of slotted content"

src/lib/expansion-panel/expansion-panel-core.ts
eliganemtyler: Sounds good to me!

src/lib/expansion-panel/expansion-panel-adapter.ts
eliganemtyler: Thanks, it looks like I misunderstood how `flatten` worked. I thought if the slotted element was for example a `<forge-scaffold>`, then `assignedElements({ flatten: true })[0] was going to return what was in the `header` slot of `<forge-scaffold>`, but I see now that's not how it works.

src/lib/expansion-panel/expansion-panel-adapter.ts
DRiFTy17: @eliganemtyler hmm ok I'll have to retest it. I thought it was overwriting my `id` upon initial render, but I must have been mistaken if you're not seeing the same thing.

src/lib/expansion-panel/expansion-panel-adapter.ts
eliganemtyler: The reason I have `document` here and not `getRootNode()` is because when `getRootNode()` gets called after component disconnect, you get a `rootNode.getElementById is not a function` error, since `getRootNode()` just finds the component itself (I guess because at this point the component is disconnected from the `document`).

The difference between expansion panel `trigger` and tooltip `anchor` in this regard is that expansion panel needs to do something with the `trigger` element (removing the event listener) on destroy. Unfortunately there isn't something like a `beforeDisconnectCallback()` hook.

What do you think about:
```
const triggerEl = deepQuerySelectorAll(document.documentElement, `#${id}`)[0] as HTMLElement;
```

Or, maybe a solution is to set a reference to the root node on init and use that instead of `this._component.getRootNode() dynamically? Is it a safe assumption (or rare enough to be ok) that the component's rootNode won't have changed since init?

DRiFTy17: If this code attempts to run while the element is disconnected I would just not do anything. It will be updated when it's connected/disconnected anyway. This is why this pattern is very finicky and can introduce memory leaks when holding onto element references that are not owned by the element itself. Once the element is disconnected from the DOM we need to drop the references (set them to undefined so the GC can clean them up) and then things like this just won't do anything until place in the DOM.

src/lib/expansion-panel/expansion-panel-adapter.ts
DRiFTy17: Yep, sorry you are right. I just retested it and all is good there. I had a separate example in my local dev demo that I copied and when I inspected this it must have been the wrong demo... All is good here.

src/lib/expansion-panel/expansion-panel-adapter.ts
eliganemtyler: Forgot to mention the other thing the expansion panel is doing when it disconnects is removing the ARIA attributes from the trigger button.

So your suggestion is to leave `const triggerEl = document.getElementById(id);` as is but add `this._adapter.triggerElementRef = undefined` to `destroy()` in core? And there will just still be that limitation with the shadow root boundaries.

eliganemtyler: Remembering now the shadow root boundary issue is kind of a no-go. It makes the `trigger` feature unusable when the expansion panel is used within another component's shadow root. What do you think about this:
```
  private _tryLocateTriggerElement(id: string): HTMLElement | null {
    if (id) {
      // If the component is still connected, find trigger in rootnode
      if (this._component.getRootNode() !== this._component) {
        return (this._component.getRootNode() as Document | ShadowRoot).getElementById(id);
        // If the component is disconnected and there is no rootnode reference, resort to querying the whole document
      } else {
        return deepQuerySelectorAll(document.documentElement, `#${id}`)[0] as HTMLElement;
      }
    } else {
      return null;
    }
  }
```
It limits the less performant `deepQuerySelectorAll` call just to when the component disconnects.

eliganemtyler:
Hey @DRiFTy17, the component is now back to largely mirroring how the tooltip manages its anchor element after our conversation. It now sets and keeps a reference to a `_triggerElement` by `trigger` `id` or `triggerElement` ref. Previously, the component would either use a `triggerElement` ref, or dynamically get the element each time by the `trigger` `id` if `triggerElement` wasn't set. The reason for this change is so the component always has a ref to the trigger element on disconnect (since previously `rootNode.getElementById()` didn't work on disconnect).

This means that the explicit `triggerElement` ref doesn't take precedence over `trigger` anymore, because the component doesn't have an obvious way of knowing if `_triggerElement` was set by `trigger` or `triggerElement`. So the property most recently set would have effect. Also like the tooltip and its anchor, it also means the component won't repeatedly check that an element still matches the `trigger` `id` -- the component doesn't know whether it's supposed to use an `id` or explicit ref. 

Overall for the sake of getting the feature out, this is at least as robust as the tooltip, even if there are better ways of handling this. Some ideas would be managing a `usingTriggerRef` state? Or managing two element refs `_triggerElementByRef` `_triggerElementById`?

The component also now sets `_triggerElement` to `undefined` on disconnect, or if `syncTrigger` runs and finds that the `_triggerElement` is disconnected (in which case it will fall back to a `trigger` `id` if one is set).

DRiFTy17 (approved):
LGTM. Thanks for all the work on this!

Once this is merged, could you then create a PR on the forge-angular repo to regenerate the Angular proxies?

PR #795, opened 2025-01-28
feat(meter): add meter component

samrichardsontylertech:
## PR Checklist

## Describe the new behavior?
This adds new Lit-based meter and key components for showing and labelling simple data.


src/lib/meter/meter/_core.scss
```diff
@@ -0,0 +1,211 @@
...
+/* stylelint-disable custom-property-pattern */
+// Manually reference the height of the meter group to avoid a circular dependency. Use half the
+// meter group's height to extend the meter behind the radius of the preceding meter.
+$leading: calc(var(--_meter-group-height) * 0.5);
+/* stylelint-enable custom-property-pattern */
+
+@mixin grouped {
+  display: block;
+  position: absolute;
+  z-index: var(--_meter-z-index);
```
DRiFTy17, line 186: Do we want to reference a token for this (and the other usages below) instead?

src/lib/meter/meter/meter.scss
```diff
@@ -0,0 +1,213 @@
...
+  .bar {
+    @include core.bar;
+    @include core.bar-animation;
+  }
+
+  &.grouped {
+    @include core.grouped;
+
+    &.muted {
+      @include core.override(color, var(--forge-theme-tertiary-container-high), value);
```
DRiFTy17, line 79: Other `var()` references in this file

src/stories/components/meter/Meter.mdx
```diff
@@ -0,0 +1,75 @@
...
+
+### Grouped
+
+Multiple related meters representing parts of a whole can be grouped together to display on the
+same track. Each meter in the group must have the same min and max values, which are managed by
+the `<forge-meter-group>` element, and can't be segmented.
+
+Grouped meters are typically accompanied by keys that label each meter with its color and
+optionally its value. Each key can be semantically associated with its meter by using a `<label>`
+element with a `for` attribute that matches the meter's `id`.
```
DRiFTy17, line 46: We should add a link to the key docs page here around the "keys" text

src/lib/meter/meter/_core.scss
```diff
@@ -0,0 +1,211 @@
...
+}
+
+@mixin vertical-bar {
+  left: initial;
+  top: 100%;
+  translate: 0 calc(-1 * var(--_meter-percentage));
+}
+
+@mixin theme($theme, $muted: false) {
+  @include override(background, var(--forge-theme-#{$theme}-container-#{if($muted, 'minimum', 'low')}), value);
```
DRiFTy17, line 92: I'm not sure we have a good pattern for this, so maybe that's why you did it this way, but we typically avoid writing `var()` references directly in favor of a mixin which generates the token. So we could use a mixin to handle generating the full `var()` expression including the fallback value. 

Same for the other usages in this file.

.storybook/preview.ts
```diff
@@ -62,6 +62,7 @@ const preview: Preview = {
       },
     },
   },
+  tags: ['autodocs']
```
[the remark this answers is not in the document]
samrichardsontylertech: Probably not, I just carried it over from the avatar branch

src/lib/meter/meter-group/meter-group.scss
```diff
@@ -0,0 +1,110 @@
...
+
+  .value {
+    @include core.value;
+  }
+
+  .track {
+    @include core.track;
+
+    ::slotted(forge-meter) {
+      position: static !important;
```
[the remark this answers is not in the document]
samrichardsontylertech: I think it used to help with positioning before I changed how that works. It can be removed now.

src/lib/key/index.scss
DRiFTy17, line 1: This file can be removed since you are already forwarding from the subdirectories

DRiFTy17 (approved):

PR #806, opened 2025-02-04
docs: fix google analytics tracking for preview pages

DRiFTy17:
## PR Checklist

## Describe the new behavior?
Fixes the Google Analytics integration by removing the embed script on the manager page and use a custom Storybook addon to integrate with Storybook page events to trigger GA integration via the `react-ga4` package.

We were previously only receiving GA data for when users navigate to the root domain of the site, but not when navigating to component stories and docs pages... This will now pick up internal navigation and push the page path to GA.

The plugin is automatically disabled when running on `localhost`.


.storybook/addons/google-analytics/register.js
```diff
@@ -0,0 +1,32 @@
+import { addons } from '@storybook/manager-api';
+import { STORY_CHANGED, STORY_ERRORED, STORY_MISSING } from '@storybook/core-events';
+import ReactGA from 'react-ga4';
+
+addons.register('storybook/google-analytics', (api) => {
+  // Do not do anything if running in localhost
+  if (globalThis.location.hostname === 'localhost') {
```
DRiFTy17, line 7: We may want to improve this check, but for now it should work fine to avoid GA events being sent from dev environments.

PR #878, opened 2025-03-22
docs(paginator): add docs about the alternative variant

DRiFTy17:


samrichardsontylertech (approved):
We should probably give this a more descriptive name than "alternative" eventually. Compact? Simple?

PR #879, opened 2025-03-22
feat(badge): refactor to use Lit

DRiFTy17:


src/lib/badge/badge.ts
```diff
@@ -47,46 +37,62 @@ declare global {
...
 
-  public get theme(): BadgeTheme {
-    return (this.getAttribute(BADGE_CONSTANTS.attributes.THEME) as BadgeTheme) ?? BADGE_CONSTANTS.defaults.THEME;
-  }
-  public set theme(value: BadgeTheme) {
-    this.setAttribute(BADGE_CONSTANTS.attributes.THEME, value);
-  }
+  /** The theme of the badge. */
+  @property({ type: String, reflect: true })
+  public theme: BadgeTheme = 'default';
```
DRiFTy17, line 65: I chose not to for this first pass, but should we consider also using CSS custom states for these common `theme` properties across components? What sort of naming convention would we use? `:state(theme-primary)` or `:state(primary)`?

samrichardsontylertech: I don't think so. I've been reserving custom states for particular things like `disabled`, `open`, etc. It feels cleaner for them to refer to behavioral changes in the components instead of purely stylistic one. We can always add states later if there's a need, anyway 🤷

samrichardsontylertech (approved):

PR #889, opened 2025-03-29
fix: enable Lit-based component compatibility with Forge `@customElement` decorator

DRiFTy17:
## PR Checklist

## Describe the new behavior?
This is a temporary change applied to all current Lit-based components in the library, and will also apply to all future Lit-based components as well. The plan is to migrate fully to Lit over time, but until we do so, we should expose the custom element tag name on Lit components via a static property so that it can be used with the Forge-based `@customElement` decorator.

This isn't necessarily to support using these decorators interchangeable within Forge (although you should be able to now), but it's to fix a compatibility issue with existing libraries out there that also make use of this decorator from the `@tylertech/forge-core` package (specifically the internal spatial library as we found compatibility issues already during testing).

While this change is not ideal, it will reduce friction as teams update Forge, but also ensures that we can continue to build components a consistent way within Forge until the time comes to fully migrate to Lit (which we will then remove these static properties).

### DIFF
--- a/src/lib/avatar/avatar.ts
+++ b/src/lib/avatar/avatar.ts
@@ -57,7 +57,7 @@ export class AvatarComponent extends LitElement implements IAvatarComponent {
   public static styles = unsafeCSS(styles);
 
   /** @deprecated Used for compatibility with legacy Forge @customElement decorator. */
-  static [CUSTOM_ELEMENT_NAME_PROPERTY] = AVATAR_TAG_NAME;
+  public static [CUSTOM_ELEMENT_NAME_PROPERTY] = AVATAR_TAG_NAME;
 
   /**
    * The text to display in the avatar.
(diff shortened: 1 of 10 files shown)


src/lib/avatar/avatar.ts
```diff
@@ -54,6 +56,9 @@ export const AVATAR_TAG_NAME: keyof HTMLElementTagNameMap = 'forge-avatar';
 export class AvatarComponent extends LitElement implements IAvatarComponent {
   public static styles = unsafeCSS(styles);
 
+  /** @deprecated Used for compatibility with legacy Forge @customElement decorator. */
+  static [CUSTOM_ELEMENT_NAME_PROPERTY] = AVATAR_TAG_NAME;
```
samrichardsontylertech, line 60: Should this have an access modifier?

DRiFTy17: It doesn't matter much to me. I suppose we might as well just make them all `private` so nobody gets the wrong idea about using them since we are intentionally introducing them as deprecated. I'll update it.

DRiFTy17: Updated. Decided to use `public` because that's actually how we intend to use them even though it's only for internal use.

samrichardsontylertech (approved):

PR #894, opened 2025-04-01
feat(app-bar): remove default auto-theming in favor of scoped CSS variables and theme attribute

DRiFTy17:
## PR Checklist

- Does this PR introduce a breaking change? Potential for visual regression, but has migration steps.

## Describe the new behavior?
This change adds a new `theme-mode="scoped"` API that allows developers to opt-out of the built-in global theme token overrides that are applied to the `<forge-app-bar>` in favor of two new tokens designed specifically for targeting certain child elements within an app bar:
- `--forge-app-bar-theme-foreground`
- `--forge-app-bar-theme-foreground-muted`

These two new tokens will not apply to any elements by default, but I have also added a new opt-in `theme="app-bar"` attribute to the following components that will inherit these new tokens when rendered within an app-bar:
- `<forge-button>`
- `<forge-icon-button>`
- `<forge-tab-bar>`

The following convenience components automatically apply the `theme="app-bar"` to their internal `<forge-icon-button>` elements for you (developers should see no change when using these components):
- `<forge-app-bar-menu-button>`
- `<forge-app-bar-profile-button>`
- `<forge-app-bar-help-button>`
- `<forge-app-bar-notification-button>`

An example of the problem that this component was facing in regards to auto-theming is demonstrated here:
https://stackblitz.com/edit/forge-v3-app-bar-theming-problem-demo?file=src%2Fcustom-app-bar-component.ts,src%2Fmain.ts

### DIFF
--- a/src/dev/src/partials/header.ejs
+++ b/src/dev/src/partials/header.ejs
@@ -1,6 +1,6 @@
 <forge-app-bar slot="header" id="page-app-bar" title-text="<%= site.title %><%= title ? ` - ${title}` : null %>" href="/">
   <forge-icon name="forge_logo" slot="logo"></forge-icon>
-  <forge-icon-button slot="end" id="dark-theme-button" theme="app-bar">
+  <forge-icon-button slot="end" id="dark-theme-button">
     <forge-icon name="brightness_3"></forge-icon>
   </forge-icon-button>
   <forge-tooltip slot="end" anchor="dark-theme-button" type="label" placement="left">Toggle theme</forge-tooltip>
--- a/src/dev/src/partials/header.ejs
+++ b/src/dev/src/partials/header.ejs
@@ -1,6 +1,6 @@
 <forge-app-bar slot="header" id="page-app-bar" title-text="<%= site.title %><%= title ? ` - ${title}` : null %>" href="/">
   <forge-icon name="forge_logo" slot="logo"></forge-icon>
-  <forge-icon-button slot="end" id="dark-theme-button">
+  <forge-icon-button slot="end" id="dark-theme-button" theme="app-bar">
     <forge-icon name="brightness_3"></forge-icon>
   </forge-icon-button>
   <forge-tooltip slot="end" anchor="dark-theme-button" type="label" placement="left">Toggle theme</forge-tooltip>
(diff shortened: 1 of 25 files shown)


samrichardsontylertech (approved):

Ross-Blakeney (approved):

nickonometry (dismissed):

src/dev/src/partials/header.ejs
```diff
@@ -1,6 +1,6 @@
 <forge-app-bar slot="header" id="page-app-bar" title-text="<%= site.title %><%= title ? ` - ${title}` : null %>" href="/">
   <forge-icon name="forge_logo" slot="logo"></forge-icon>
-  <forge-icon-button slot="end" id="dark-theme-button">
+  <forge-icon-button slot="end" id="dark-theme-button" theme="app-bar">
```
MikeMatusz, line 3: While this migration step is simple, it's still unexpected in a minor version, and if I weren't following Forge updates so closely, I'd be frustrated having to track it down, even if it was noted in the changelog.  A team could merge a Dependabot PR and end up with a regression.  I know of some teams that are gun-shy about taking even patch updates of Forge due to past issues, and while I feel like that situation has improved, it's just not a great developer experience for a sem-ver library to break existing usages outside of major versions.

I definitely understand how the current behavior is problematic, and that Forge can't just release major versions all the time, and maybe this does only affect a small number of applications (though the first two I checked here will both be affected).  I also see where making this opt-in to start has extra overhead, but I feel like it's the better choice for a library we want developers to trust.  Yes, if anyone misses the memo to enable it, they could run into the problem it fixes, but finding an issue and learning it already has a solution is a better experience than having something become a problem that wasn't before.

DRiFTy17: @MikeMatusz This is exactly my concern with it. I'd like to tread very lightly with this one to avoid any disruption, hence wanting to gather feedback on it beforehand so I appreciate your thoughts on it!

Let me see what an opt-in scenario would look like. I think we might be able to find a hybrid middle ground that might be the best path forward for now.

MikeMatusz (approved):

tyl-paul-lapczynski (approved):

PR #904, opened 2025-04-06
fix(expansion-panel): fix connected check

### LINKED ISSUE #903: [expansion-panel] component may not be connected before getTriggerById
**Describe the bug:**
In an Angular app, it's possible for the incremental rendering to cause the expansion-panel to initialize before the component is connected to the document, so `this._component.getRootNode().getElementById(...)` errors, as the component's root node is itself, which does not have `getElementById(...)`

The fix is just to move the `setTriggerElementById` call to inside the `if (this._adapter.isConnected)` block in `set trigger`

**To Reproduce:**
1. Create an angular app with forge 3.8.0 and 
2. In a component that will render unconditionally, create an expansion panel with trigger, e.g.
```
<forge-expansion-panel trigger="hello">
  <div id="hello" slot="header">header</div>
  <div>content</div>
</forge-expansion-panel>
```

**Expected behavior:**
The expansion panel should not reference its root node before it's connected.

**Screenshots:**
Current:
Image
Image
Image
Image

With fix:
Image

 - Forge version: 2.8.0
 - Browser: All (tested on Chrome)
 - Platform: All (tested on desktop)
 - OS: All (tested on Windows)

**Additional context:**

eliganemtyler:
## PR Checklist

  Fixes #903 

## Describe the new behavior?
Oops! This call to `_getTriggerElementById` needed to happen inside the `isConnected` check to be compatible with Angular's incremental rendering.

Not sure if there's a way to write a test for this common Angular-specific gotcha. The component passes current tests before and after this fix.


eliganemtyler:
Looks like I don't have the ability to link the issue but I described it here: https://github.com/tyler-technologies-oss/forge/pull/904

DRiFTy17 (approved):

PR #912, opened 2025-04-16
chore: consolidating various defineXComponent calls to use the standard forge-core defineCustomElement call

bctt:
## PR Checklist

## Describe the new behavior?
Modifying a handful of defineXComponent calls in some component `index.ts` files to use the standard `defineCustomElement` call that `forge-core` offers.

### DIFF
--- a/src/lib/avatar/index.ts
+++ b/src/lib/avatar/index.ts
@@ -1,9 +1,9 @@
-import { tryDefine } from '@tylertech/forge-core';
-import { AVATAR_TAG_NAME, AvatarComponent } from './avatar';
+import { defineCustomElement } from '@tylertech/forge-core';
+import { AvatarComponent } from './avatar';
 
 export * from './avatar';
 export * from './avatar-constants';
 
 export function defineAvatarComponent(): void {
-  tryDefine(AVATAR_TAG_NAME, AvatarComponent);
+  defineCustomElement(AvatarComponent);
 }
--- a/src/lib/badge/index.ts
+++ b/src/lib/badge/index.ts
@@ -1,10 +1,10 @@
-import { tryDefine } from '@tylertech/forge-core';
-import { BADGE_TAG_NAME, BadgeComponent } from './badge';
+import { defineCustomElement } from '@tylertech/forge-core';
+import { BadgeComponent } from './badge';
 
 export * from './badge-component-delegate';
 export * from './badge-constants';
 export * from './badge';
 
 export function defineBadgeComponent(): void {
-  tryDefine(BADGE_TAG_NAME, BadgeComponent);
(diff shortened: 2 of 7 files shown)


DRiFTy17:
@bctt Thanks for the PR! I'm just curious if there is a specific problem that this change solves? The reason I ask is because we only just recently made the opposite of this change, and it's only for our Lit-based components.

We use `tryDefine()` for the Lit-based components so it can gracefully handle potentially executing the function more than once. Ideally we do not attempt to define custom elements more than once of course, but this is just for backwards compatibility and consistency with the rest of legacy Forge.

Thanks again, I just want to better understand the motivation behind this one before the next steps.

bctt:
Thanks for looking this over @DRiFTy17 ! The main drive behind this change was to have consistency across all of the defineComponent() calls in Forge. 

After digging through `forge-core`, I now have a better understanding of how `forge` declares its components with the `CustomComponentRegistry`. 
`defineCustomComponent` is already making a call to `tryDefine`, but it passes along the component's `HTMLElementTagNameMap` that is built into the component instead of taking in a different name as a parameter, like what is seen with `forge-badge` in the current version of `badge/index.ts`

I may be missing some knowledge if this change is a regression for those components though... was there somehow something that `tryDefine` was accomplishing that was more concise than what `defineCustomComponent` was doing with its logic that included `tryDefine`?

DRiFTy17:
You're not missing anything, we're just in the middle of transition period right now between our legacy vanilla-based Web Component architecture and the new Lit-based components that everything will be eventually migrated to over time. We had a bug we needed to fix in the Lit-based components related to `defineCustomElement()` and we swapped to calling `tryDefine()` directly. We then came back recently in #889 to fix compatibility with the `dependencies` property in the Forge-based `@customElement` decorator, which subsequently also fixed the original usage of `defineCustomElement()`...

Sorry, I should have also provided a link to the change I mentioned above where this occurred: https://github.com/tyler-technologies-oss/forge/pull/889

That PR technically enables us to use `defineCustomElement()` again because it adds the missing static property to the Lit component classes (which is used to pass to `tryDefine()` as seen [here](https://github.com/tyler-technologies-oss/forge-core/blob/main/src/custom-elements/component-utils.ts#L16)).

A long term goal for Forge is to eliminate the need for these "definition functions" and static properties in favor of just side-effect imports like this for example:
```javascript
import '@tylertech/forge/button';
```

Long story short, I think your change will be fine to revert back to `defineCustomElement()` for consistency but I just wanted to give you the full story of how it got in this state (sorry again for the confusion!). We'll need to do some testing on this by using in a separate app before releasing but that's about it!

DRiFTy17 (approved):

PR #965, opened 2025-07-30
fix(field): add overflow and isolation token to the inner container element

nickonometry:
## PR Checklist

- Tests for the changes have been added/updated: [N]

## Describe the new behavior?
There's situations, especially when using 3rd party plugins with the text field, that we need to modify the overflow and isolation CSS properties. This change gives us an easy way to do that, but leaves the default values

image


DRiFTy17 (approved):
I'm fine with exposing these, but if anyone adjusts these styles while using the inset label, things like likely not work as expected in terms of the label animation and auto-fill styles. Otherwise it should be fine for labels outside of the field.

PR #1005, opened 2025-10-22
feat(label-value): refactor to use lit

jasonsubers-tyler:
## PR Checklist

## Describe the new behavior?
Refactored the "Label Value" component to use Lit

## Additional information
Retained deprecated "dense" prop, if used will map to "inline" via willUpdate() call


src/lib/label-value/label-value.ts
```diff
@@ -55,81 +53,71 @@ declare global {
...
+  public dense = false;
+
+  public override willUpdate(changedProperties: PropertyValues<this>): void {
+    if (changedProperties.has('dense')) {
+      if (this.hasUpdated || this.hasAttribute('dense')) {
+        // Ignore the initialization cycle. On subsequent updates, map dense to inline
+        if (this.inline !== this.dense) {
+          this.inline = this.dense;
+        }
+      }
```
DRiFTy17: I know it's unfortunate that we have to keep these properties in sync for now to avoid breaking changes, but I'm not sure how I feel about this block. I also noticed that the dense attribute does not get removed if setting inline to false. Not really a big deal since typically only one of these properties is in use at a time, but better to handle it gracefully. Maybe we can clean this up and make it more simple/readable?

PR #1006, opened 2025-10-24
feat(backdrop): refactor to use lit

jasonsubers-tyler:
## PR Checklist

## Describe the new behavior?
Refactored the "Backdrop" component to use Lit

## Additional information
* In fade in/fade out test runner, "animationend" event never fires, likely due to animation collapse in test environment. As a result, fade in/fade out tests assert on visible attribute only; removed class assertions.

* Minor style fix for "Label-Value"

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -23787,9 +23787,9 @@
       }
     },
     "node_modules/vite": {
-      "version": "7.1.6",
-      "resolved": "https://registry.npmjs.org/vite/-/vite-7.1.6.tgz",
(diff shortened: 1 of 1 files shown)


DRiFTy17:
It looks like you probably branched form your label-value branch, so the changes for that component are also included here. Could you fix that so it only includes your backdrop changes? I know you mentioned you made a small styling fix to the label value, but that should be included in the other branch or separate on its own so we can track these changes more easily.

PR #1012, opened 2025-10-30
feat(paginator): add interaction methods for convenience

### LINKED ISSUE #1011: [paginator] expose actions buttons as properties on component
**Describe the feature:**
Currently, if you want to write automated tests using tools such as Selenium or Playwright, you must use the `shadowRoot` property to perform clicks on the `forge-paginator` action buttons. It would be helpful if the first, previous, next, and last action icon buttons were exposed as properties on the paginator component. This would allow easier access to them for automation purposes. We may also want to expose the page size options select at the same time.

I thought about if it would make sense to expose these as slotted elements outside the shadow DOM. Similar to the `forge-date-picker` where it adds a slotted icon button for the calendar icon automatically, but it's accessible outside the shadow DOM. Because this component is so self contained, I don't think that would make sense and may convolute the component's current API, but I did want to mention it as another alternative.

**Additional context:**
N/A

DRiFTy17:
## PR Checklist

## Describe the new behavior?
Adds public navigation methods to enable external testing tools to interact with paginator without accessing shadow DOM internals.

- `goToFirstPage()`, `goToPreviousPage()`, `goToNextPage()`, `goToLastPage()` - Navigation actions
- `canGoToFirstPage()`, `canGoToPreviousPage()`, `canGoToNextPage()`, `canGoToLastPage()` - State access/validation

## Additional information
Resolves #1011

### DIFF
--- a/src/lib/paginator/paginator-constants.ts
+++ b/src/lib/paginator/paginator-constants.ts
@@ -77,13 +77,3 @@ export interface IPaginatorRangeState {
 }
 
 export type PaginatorRangeLabelBuilder = ((state: IPaginatorRangeState) => string) | undefined | null;
-
-export interface IPaginatorPageInfo {
-  pageIndex: number;
-  pageSize: number;
-  total: number;
-  offset: number;
-  pageStart: number;
-  pageEnd: number;
-  totalPages: number;
-}


src/lib/paginator/paginator-constants.ts
```diff
@@ -77,3 +77,13 @@ export interface IPaginatorRangeState {
...
+
+export interface IPaginatorPageInfo {
+  pageIndex: number;
+  pageSize: number;
+  total: number;
+  offset: number;
+  pageStart: number;
+  pageEnd: number;
+  totalPages: number;
+}
```
alienriver49: Do we still need this?

DRiFTy17: Nope! Good catch, I was going to expose the state as a single method but chose to revert that since we have individual properties for everything already. Updated, thanks.

PR #1040, opened 2025-12-15
feat(autocomplete): add select first option on blur

### LINKED ISSUE #1039: [autocomplete] Allow the first available option to be selected automatically when blurring mid-filter
**Describe the feature:**

We have clients that type really fast, sometimes faster than our application server or the filter function can run, and they navigate almost exclusively with keyboard. With the current autocomplete—if you type faster than the debounce or filter promise (i.e. the loading indicator is visible) and then tab out of the component, nothing is selected. (You can recreate it if you're fast enough on the forge-autocomplete demo in storybook).

A lot of users expect that doing this would instead select the first result, like it does when you have filterFocusFirst set and the filter is complete. Especially since this was the behavior of these sort of inputs in our legacy application they’re migrating from.

We’d need some sort of a `select-first-option-on-blur` attribute for when the user is faster is than the debounce + filter promise.

We tried hooking/overriding the autocomplete internals directly, but had side effects that were unacceptable. We also considered creating a custom autocomplete in angular using other forge components, but we'd prefer to not have two different versions of an autocomplete in our codebase. Right now, we are using `pnpm` to patch forge to get this behavior.

**Additional context:**

This was mentioned in a service desk ticket with CorpDev support. I already have a branch ready on my fork of forge that implements this feature.

jaxcksn:
## PR Checklist

## Describe the new behavior?

Adds a property that enables the first available option that matches to be selected automatically when blurring mid-filter. It will close the dropdown and then select when the filter promise resolves. It is disabled by default, and does not work when the input mode is stateless.
 
## Additional information

Resolves #1039

### DIFF
--- a/src/lib/paginator/paginator-adapter.ts
+++ b/src/lib/paginator/paginator-adapter.ts
@@ -161,19 +161,16 @@ export class PaginatorAdapter extends BaseAdapter<IPaginatorComponent> implement
   }
 
   public setFirstPageButtonEnabled(enabled: boolean): void {
-    this._firstPageButton.disabled = !enabled;
+    this._setButtonEnabled(this._firstPageButton, enabled);
   }
-
   public setPreviousPageButtonEnabled(enabled: boolean): void {
-    this._previousPageButton.disabled = !enabled;
+    this._setButtonEnabled(this._previousPageButton, enabled);
   }
-
   public setNextPageButtonEnabled(enabled: boolean): void {
-    this._nextPageButton.disabled = !enabled;
+    this._setButtonEnabled(this._nextPageButton, enabled);
   }
-
   public setLastPageButtonEnabled(enabled: boolean): void {
-    this._lastPageButton.disabled = !enabled;
+    this._setButtonEnabled(this._lastPageButton, enabled);
   }
 
   public setPageSizeSelectEnabled(enabled: boolean): void {
@@ -205,6 +202,17 @@ export class PaginatorAdapter extends BaseAdapter<IPaginatorComponent> implement
         el.focus({ ...options, preventScroll });
         return;
       }
+    }
+  }
+
+  private _setButtonEnabled(btn: HTMLElement, enabled: boolean): void {
+    if (enabled) {
+      btn.removeAttribute('disabled');
+      btn.setAttribute('tabindex', '0');
+    } else {
+      btn.setAttribute('disabled', 'true');
+      btn.setAttribute('tabindex', '-1');
+      btn.blur();
     }
   }
 }


src/lib/autocomplete/autocomplete-core.ts
```diff
@@ -239,12 +242,22 @@ export class AutocompleteCore extends ListDropdownAwareCore implements IAutocomp
...
       this._closeDropdown();
     }
 
     // If we are in stateless mode, we don't need to do anything further
     if (this._mode === AutocompleteMode.Stateless) {
+      this._selectOnBlurPending = false;
+      return;
+    }
+
+    if (shouldSelectFirstOptionOnBlur) {
```
MikeMatusz: Couldn't this be simplified?  If it's Stateless, then shouldSelectFirstOnBlur() will return false, so I'm not sure explicitly setting it is necessary.

```ts
if (this._isDropdownOpen) {
   this._closeDropdown();
}

this._selectOnBlurPending = this._shouldSelectFirstOnBlur();
if (this._selectOnBlurPending || this._mode === AutocompleteMode.Stateless) {
    return;
}
```

src/lib/autocomplete/autocomplete-core.ts
MikeMatusz: Do we need to worry about clearing `_selectOnBlurPending` if the control receives focus again before the original filter promise completes?  Definitely an edge case, but I was able to type, tab off, tab back (maybe because I made a typo), and start typing again.  I actually see the dropdown appear for the original filterText then it selects the first matching value for the updated filterText without showing the updated filter.

src/lib/autocomplete/autocomplete-core.ts
```diff
@@ -258,6 +271,17 @@ export class AutocompleteCore extends ListDropdownAwareCore implements IAutocomp
...
   }
 
+  private _shouldSelectFirstOptionOnBlur(): boolean {
+    return (
+      this._selectFirstOptionOnBlur &&
+      !this._multiple &&
+      this._mode !== AutocompleteMode.Stateless &&
+      !!this._filter &&
+      !!this._filterText &&
+      !this._selectedOptions.length
```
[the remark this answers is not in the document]
jaxcksn: Nope! This was unintended. I've removed that last condition from the behavior.

src/lib/autocomplete/autocomplete-core.ts
```diff
@@ -399,22 +412,42 @@ export class AutocompleteCore extends ListDropdownAwareCore implements IAutocomp
...
+
+    if (!this._options.length) {
+      this._selectOnBlurPending = false;
       if (this._isDropdownOpen) {
         this._closeDropdown();
       }
+      if (shouldSelectOnBlur && !this._allowUnmatched) {
+        this._filterText = '';
+        this._adapter.setSelectedText('');
+      }
```
MikeMatusz: I didn't catch this originally because it didn't support updating when a value was already selected, but looks like we might need to additionally call `_clearValue()` or something similar.  I notice that if a value is selected, and I type a string with no matches and tab off, the filter text clears, but `forge-autocomplete-change` never emits and the "Selected Value" in the demo app isn't cleared.  Aside from that, though, all the updates look good and I'm excited about this feature!

PR #1078, opened 2026-02-25
feat(icon): migrate to lit

samrichardsontylertech:
## Summary
Migrated the icon component to Lit.

## Checklist
- [x] Tests added/updated

## Breaking Changes
The `IIconProperties` interface was changed to more accurately reflect the properties in the component class, making members that may be undefined in the class optional.


packages/forge/src/lib/icon/index.ts
DRiFTy17: Mark as deprecated in favor of side effect import

Issue #1103, opened 2026-03-06
[text-field] `for` `id` pair doesn't work asynchronously


eliganemtyler:
### Package

forge

### Version

3,13.1

### Description

In the case where `<forge-text-field>` is connected to DOM asynchronously and `slot="label"` is explicitly set on the `<label>`,  the `<input>` and `<label>` will not be associated by a `for` `id` pair as expected.

This is also relevant to the `TextFieldComponentDelegate` where `slot="label"` is explicitly set (and often used in async contexts)

### Steps to Reproduce

1. Create a `<forge-text-field>` with a child with `slot="label"` explicitly defined.
```
const textField = document.createElement('forge-text-field');
const label = document.createElement('label');
const input = document.createElement('input');
label.innerText = 'mylabel';
input.type = 'text';
label.slot = 'label'
textField.appendChild(label);
textField.appendChild(input);
```
OR
```
const textField = new TextFieldComponentDelegate({options: {label: 'my label'}})
```
2. Attach it to the dom asynchronously.
```
const myDiv = document.querySelector('#mydiv');
setTimeout(() => {
  myDiv.appendChild(textField)
}, 0)
```
3. The resulting `<input>` and `<label>` are not associated by `for` and `id`.

### Expected Behavior

The `<input>` and `<label>` are always correctly associated when connected to the DOM regardless of timing and explicit `slot` assignment.

PR #1104, opened 2026-03-06
Fix/text field forid async

eliganemtyler:
## Summary
Fixes https://github.com/tyler-technologies-oss/forge/issues/1103

In the case where `<forge-text-field>` is connected to the DOM asynchronously with explicit `slot="label"`, the `slotchange` event that triggers `tryConnectSlottedLabel()` doesn't fire. `_initializeSlots()` should cover this edge case, but `getAllSlotElements()` returned the label `slot` before the input `slot`, which meant the `<input>` ref wasn't ready in time for `tryConnectSlottedLabel()`. Now `tryConnectSlottedLabel()` gets called on default/input slot change which means the `getAllSlotElements()` array order doesn't matter.

Maybe this is incidentally also fixing the edge case (that practically probably never happens) where the `<input>` is swapped out dynamically?

## Checklist
- [x] Tests added/updated

## Breaking Changes
None

### DIFF
--- a/.changeset/lucky-dancers-reply.md
+++ b/.changeset/lucky-dancers-reply.md
@@ -1,0 +1,5 @@
+---
+'@tylertech/forge': patch
+---
+
+fix(menu): expand cascade menu hover area to full list-item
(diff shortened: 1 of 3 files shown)


.changeset/tricky-crabs-beg.md
```diff
@@ -0,0 +1,5 @@
+---
+'@tylertech/forge': patch
+---
+
+Fix missing for-id pair in async edge case
```
DRiFTy17: We've been using convention commit format for the changesets so that it reads better in the changelog. See the release PR here for examples:
https://github.com/tyler-technologies-oss/forge/pull/1068

So something like this:
`fix(text-field): fix missing for-id pair in async edge case`

PR #1139, opened 2026-04-28
feat(tabs): migrate to lit

### LINKED ISSUE #1155: [tab-bar] active-tab does not work when tabs are added before the tab-bar has been connected to the DOM
### Package

forge

### Version

3.14.3

### Description
Appears to be related to #1154. Recommend doing an audit on these slot listeners to make sure they synchronize what's necessary when becoming connected, OR, register slot listeners before connected if possible.

I have an Angular app where the forge-tab-bar has [activeIndex] properly bound, but the active tab is not showing the active indicator. IN debugging, you can actually see `set activeTab` firing for the active tab before tabs exist. It then assigns `this._activeTab` even though you don't have tabs. When tabs do get set, `set activeTab` does appear to get called again, but won't actually add the selected attribute because `_activeTab === value` and it returns out. There is a syncTabState designed to do this, but it won't work if the forge-tab-bar hasn't been connected yet when the tabs are added separately.

This is a common workflow to create elements in this nature with frameworks such as angular and react, and absolutely should be supported.

Image

### Steps to Reproduce

1. Go to forge docs for tab-bar https://forge.tylerdev.io/forge/v3/?path=/docs/components-tabs--docs
2. Inspect the root-inner div that the forge-tab-bar is in and keep it selected
3. In dev tools console do
```
const div = document.createElement('div');
div.innerHTML = '<forge-tab-bar data-aria-label="Demo tabs" active-tab="0"></forge-tab-bar>';
div.children[0].appendChild(document.createElement('forge-tab'));
div.children[0].children[0].innerText = 'Test';
```
then
```
$0.appendChild(div);
```

### Expected Behavior

The activeTab should always be synced from the forge-tab-bar to the forge tabs. It appears that there is code to do this `syncTabState` but that only fires from a slot listener, which isn't set until the tab bar has been connected. It seems that this state should always be synced. Perhaps the slot listener should be registered before being connected (if possible), else, when it does get connected, listeners need to fire that are necessary to sync state because they missed it.

### Environment

Chrome, Windows

samrichardsontylertech:
## Summary
Migrated the tab and tab bar components to Lit.

Also added new features and fixes:
- ~Deprecated the `activeTab` property in favor of new `activeTabIndex`, `activeTabName`, and `activeTabElement` properties~
- ~Tabs accept and can be tracked by a `name` attribute~
- Tabs can be removed with the `backspace` or `delete` key
  - Enabled with a new `closable` property
  - Dispatches a cancelable `forge-tab-close` event
- Added initial support for opening popover menus from tabs with `Shift+F10`
- Borders appear on the scroll buttons when there is overflow in that direction
  - Added `overflow-divider-width` and `overflow-divider-color` tokens
- Replaced host attributes in the CSS with custom state selectors and internal classes to prepare for attribute reflection to be deprecated
- Properties are passed from the tab bar to its child tabs with Lit context
- Created a reusable focus group class wth Lit directive and reactive controller implementations
- Refactored keyboard shortcuts to use a `KeyActionController` from the core utils
- Fixed scroll detection and focus management bugs
- Added support for native nearest container scrolling as a progressive enhancement
- Tab management logic has been moved to the tab bar component
- Rewrote `isDefined` in Forge Core to be a user-defined type guard

## Checklist
- [x] Tests added/updated

## Breaking Changes
None


packages/forge/src/lib/tabs/tab/tab.ts
```diff
@@ -1,37 +1,33 @@
-import { customElement, attachShadowTemplate, coerceBoolean, coreProperty } from '@tylertech/forge-core';
-
-import { TabAdapter } from './tab-adapter.js';
-import { TabCore } from './tab-core.js';
-import { TAB_CONSTANTS } from './tab-constants.js';
-import { BaseComponent, IBaseComponent } from '../../core/base/base-component.js';
+import { consume } from '@lit/context';
+import { CUSTOM_ELEMENT_DEPENDENCIES_PROPERTY, CUSTOM_ELEMENT_NAME_PROPERTY } from '@tylertech/forge-core';
+import { setDefaultAria } from '@tylertech/forge/core/utils/a11y-utils.js';
```
DRiFTy17: This should be a relative import path

packages/forge/src/lib/tabs/tab/tab.ts
```diff
@@ -100,75 +83,258 @@ declare global {
...
-    this._core.setFocus(options);
+    if (options?.focusVisible && this._focusIndicator) {
+      this._focusIndicator.active = true;
+    }
+  }
+
+  /**
+   * Removes the tab.
+   */
+  public remove(): void {
```
DRiFTy17: Reminder to rename this method to avoid shadowing. Maybe `requestRemove()`?

packages/forge/src/lib/tabs/tab-bar/tab-bar.scss
```diff
@@ -86,13 +81,13 @@ forge-icon-button {
...
-  .forge-tab-bar {
-    @include override(justify, flex-start, value);
-    @include override(stretch, 0, value);
-  }
+.forge-tab-bar.clustered {
+  @include override(justify, flex-start, value);
+  @include override(stretch, 0, value);
 }
 
+// TODO: `clustered is a boolean attribute/property -- is anyone actually setting a value to it? If not, remove these. If so, change the property.
```
DRiFTy17, line 89: Yeah this is a strange undocumented feature. My guess is that it was implemented one way, and was only half removed or something when changed. I don't know how often it is used, if it all. Not really a pattern we want to promote Maybe we need to update the type to be `boolean | TabBarClusteredType` to support both for backwards compat?

packages/forge/src/lib/tabs/tab/tab-constants.ts
```diff
@@ -22,17 +22,21 @@ const selectors = {
...
 
 const strings = {
   EASING: 'cubic-bezier(0.4, 0, 0.2, 1)'
 };
 
 const numbers = {
-  ANIMATION_DURATION: 250
+  ANIMATION_DURATION: 250,
+  TAB_INDEX_ACTIVE: 0,
+  TAB_INDEX_INACTIVE: -1
```
DRiFTy17: I'm not sure it's worth having these as constants?

packages/forge/src/lib/tabs/tab/tab.ts
```diff
@@ -100,75 +83,258 @@ declare global {
...
+  }
+}
+
+declare global {
+  interface HTMLElementTagNameMap {
+    'forge-tab': ITabComponent;
+  }
+
+  interface HTMLElementEventMap {
+    'forge-tab-select': CustomEvent<void>;
```
DRiFTy17, line 341: We should also add the `forge-tab-request-remove` event here?

packages/forge/src/lib/tabs/tab/tab.ts
```diff
@@ -100,75 +83,258 @@ declare global {
...
+   */
+  public remove(): void {
+    const event = new Event(TAB_CONSTANTS.events.REQUEST_REMOVE, { composed: true, bubbles: true });
+    this.dispatchEvent(event);
+  }
+
+  /**
+   * Plays the state layer animation.
+   */
+  public playStateLayerAnimation(): void {
```
DRiFTy17: This feels like an "internal" method, although I know we need to use from the tab bar. Maybe we should use a symbol method instead to avoid it being "seen/used" as part of the public API?

Super minor of course, but just a precedent/pattern that we're setting for future changes.

packages/forge/src/lib/core/utils/key-action.ts
```diff
@@ -108,19 +108,19 @@ export class KeyActionController implements ReactiveController {
   }
 
   public hostConnected(): void {
-    (this.host as ReactiveElement).addEventListener('keydown', this._keyDownListener);
+    (this.host as ReactiveElement).addEventListener('keydown', this.#keyDownListener);
```
DRiFTy17, line 118: Should the event type be configurable? This is related to somewhat recent discussions around when to use `keydown` vs `keyup` to mimic native behavior, and it seems that `keyup` is desirable at times.

DRiFTy17 (commented):
Just an initial review. Nothing major jumped out at me. This refactor looks well thought out, and the conversion to Lit with context, controllers, and directives is great. Will dig in further once the tests are passing and it's out of draft, and then I will focus more on usage validation at that point to ensure backwards compatibility + stability.

packages/forge/src/lib/tabs/tab/tab.ts
samrichardsontylertech: I'm just removing the method altogether. The tab bar already handles this well enough through the slot change listener.

packages/forge/src/lib/tabs/tab-bar/tab-bar.scss
samrichardsontylertech: I'd prefer to handle the justification with just CSS but we might be stuck with this anyway

packages/forge/src/lib/tabs/tab/tab-constants.ts
samrichardsontylertech: Thank Claude!

packages/forge/src/lib/tabs/tab/tab.ts
samrichardsontylertech: Just altogether removed that one 👍

packages/forge/src/lib/tabs/tab/tab.ts
samrichardsontylertech: You're right, I'm just starting to feel like symbol methods are messy. It's not a big deal though.

packages/forge/src/lib/core/utils/key-action.ts
samrichardsontylertech: The key actions controller uses the `repeat` property to do basically the same thing as `keyup`. I think this is more consistent with native behavior but I'm open to discussing it.

samrichardsontylertech:
The latest commit fixes a circular reference error by moving the tab bar context constants to the constants file, but for some reason this seems to have broken the actual usage of the contexts.

DRiFTy17 (commented):
This is looking really good! Just a few questions:

- What should happen when the selected tab is removed from the DOM?
    - Should we lose `selectedIndex` and `selectedTab` or should they be preserved? Controlled (consumer) vs uncontrolled (component)
- Should tabs support dynamic `name` changes?
    - This could also be a timing concern with framework bindings updating `name` after `<forge-tab>` is connected. Should selection state sync at this time?
- How should newly added tabs be handled if they match the `selectedTab` name? It's likely that `selectedTab` would be set before `<forge-tab>` elements are connected. There appears to be some conflict if `selectedTab` is set, then a `<forge-tab>` is added, it will auto-select that tab even if it doesn't match the name of the `selectedTab`.
- Should enter key activate scroll buttons?

DRiFTy17 (commented):
I'm seeing an error when viewing the docs page in Storybook:
https://forge.tylerdev.io/forge/pr-1139/?path=/docs/components-tabs--docs

Image

packages/forge/src/dev/pages/tabs/tabs.ejs
DRiFTy17, line 1: We should probably remove the `name` attributes here for now until we get the new name-based feature out to avoid confusion.

DRiFTy17 (dismissed):

PR #1144, opened 2026-05-04
feat: global config decorator

samrichardsontylertech:
## Summary
Created a global config decorator to replace the adapter-based approach currently used. The decorator is compatible with the new Lit architecture, keeps value definitions in one place, and solves issues with global configuration values overriding externally set values when a component initializes.

## Checklist
- [ ] Tests added/updated

## Breaking Changes
None

### DIFF
--- a/packages/forge/src/lib/core/decorators/global-configuration-decorator.ts
+++ b/packages/forge/src/lib/core/decorators/global-configuration-decorator.ts
@@ -10,38 +10,39 @@ interface CoreWithAdapter {
  *
  * When the property is first accessed, the decorator will look up the tag name and property name
  * in the global configuration object and set the initial value from it. If not found in the
- * global configuration, it will use the provided default value.
+ * global configuration, it will use the property's initial value as the default.
  *
  * Works with both HTMLElement instances (Lit components) and Core classes (legacy pattern).
  * For Core classes, requires an `_adapter` property that provides access to the host element.
  *
- * @param defaultValue The default value to use if no global configuration is found
- * @throws {TypeError} If the default value type doesn't match the property type
+ * @throws {TypeError} If the global configuration value type doesn't match the property's initial value type
  *
  * @example
  * ```typescript
  * // On a Lit component
- * @globalConfig(false)
- * public someProperty: boolean;
+ * @globalConfig()
+ * public someProperty: boolean = false;
  *
  * // On a Core class
- * @globalConfig('outlined')
- * private _variant!: FieldVariant;
+ * @globalConfig()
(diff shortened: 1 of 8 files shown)


packages/forge/src/lib/field/field-core.ts
```diff
@@ -62,8 +65,16 @@ export class FieldCore implements IFieldCore {
   public initialize(): void {
     this._adapter.addRootListener('slotchange', this._slotChangeListener);
     this._adapter.initializeSlots();
-    this._adapter.tryApplyGlobalConfiguration(['labelPosition', 'variant']);
+
+    // Setting the label position and variant here for global config decorator compatibility
+    // Skipping if there are already host attributes set to avoid overwriting values that may have been set externally
     this._adapter.setLabelPosition(this._labelPosition);
+    if (!this._adapter.getHostAttribute(FIELD_CONSTANTS.attributes.VARIANT)) {
```
DRiFTy17: Shouldn't this reference `LABEL_POSITION` instead of `VARIANT`?

packages/forge/src/lib/field/field-core.ts
```diff
@@ -62,8 +65,16 @@ export class FieldCore implements IFieldCore {
...
+      this._adapter.setHostAttribute(FIELD_CONSTANTS.attributes.LABEL_POSITION, this._labelPosition);
```
DRiFTy17, line 74: Doesn't this mean that every time the component initializes, it will sprout a host attribute if one is not set? I wouldn't think we would want to introduce that?

packages/forge/src/lib/core/decorators/global-configuration-decorator.ts
```diff
@@ -0,0 +1,94 @@
...
+ * // On a Lit component
+ * @globalConfig(false)
+ * public someProperty: boolean;
+ *
+ * // On a Core class
+ * @globalConfig('outlined')
+ * private _variant!: FieldVariant;
+ * ```
+ */
+export function globalConfig<T>(defaultValue: T): PropertyDecorator {
```
DRiFTy17: We should add some tests for usage of this decorator.

packages/forge/src/lib/field/field-core.ts
samrichardsontylertech: Yeah, it's a side effect of how we're styling the components. The default label position/theme is targeted by the lack of the attribute and the others depend on the attribute being there. Our Lit components don't have this problem because the attributes aren't used in the CSS.

I can get around this now by also checking whether the value differs from the default though.

DRiFTy17 (approved):

PR #1145, opened 2026-05-05
feat(blocks): Forge Blocks — a library of reusable HTML block templates

nickonometry:
## Summary

**Forge Blocks** — a standalone package of reusable, standalone HTML templates demonstrating Tyler Forge component patterns (`blocks/`).

- **Block format & compiler** — blocks are plain HTML files with a `@block`/`@description`/`@tags`/`@type` metadata comment, compiled into a shared base layout (`src/includes/base.html`) via Handlebars, with support for reusable partials (`src/partials/*.hbs`).
- **Dev server** — a Vite plugin (`vite-plugin.ts`) serves blocks on the fly, rewrites block URLs, serves a live `/manifest.json`, and hot-reloads on partial/index changes. Added a root-level `pnpm dev:blocks` script to launch it from the repo root.
- **Build pipeline** — `build-blocks.ts` compiles every block to `dist/`, copies static assets, and generates the manifest for production.
- **Manifest generation** — `generate-manifest.ts` discovers all blocks, extracts metadata, auto-detects which `forge-*` components each block uses (including through nested partials), and validates required metadata annotations.
- **Screenshots** — `generate-screenshots.ts` uses Playwright + Vite to capture each block, then resizes/optimizes to WebP thumbnails via `sharp`, with incremental regeneration (only rebuilds screenshots for changed blocks unless `--force`).
- **Content** — ~100+ new blocks across application layout, cards, tables, forms, and component-demo categories, largely migrated from the existing Storybook stories plus a handful of new patterns (context switchers, table with detail sidesheet, complex forms, cohort status card, etc.).
- **CI/CD** — new GitHub Actions workflows for blocks PR previews (`blocks-preview.yml` + cleanup) and production deploy (`deploy-blocks.yml`), plus updates to the Storybook build/deploy workflows so blocks ship alongside it.

## Checklist
- [ ] Tests added/updated
- [x] Docs updated (`blocks/CLAUDE.md`, `blocks/README.md`, root `CLAUDE.md`)
- [ ] Changeset added (`pnpm changeset`) — n/a, `blocks` is a private, unpublished package

## Breaking Changes
None — new package, no changes to published `@tylertech/forge`/`@tylertech/forge-extended` APIs.


blocks/scripts/generate-manifest.ts
```diff
@@ -0,0 +1,66 @@
...
+  const blockMatch = match[1]?.match(/@block\s+(.+)/);
+  const descMatch = match[2]?.match(/@description\s+(.+)/);
+  const tagsMatch = match[3]?.match(/@tags\s+(.+)/);
+
+  const name = blockMatch?.[1]?.trim() || path.basename(filePath, '.html');
+  const description = descMatch?.[1]?.trim() || '';
+  const tags = tagsMatch?.[1]?.split(',').map(t => t.trim()).filter(Boolean) || [];
+
+  return { name, description, tags };
+}
```
DRiFTy17: This seems to be duplicated across this file and `vite.config.ts`, should probably extract to a shared module along with the `METADATA_REGEX` and `BlockMetadata`?

DRiFTy17:
Can you add a concise README as well?

blocks/src/scripts/vite-plugin.ts
```diff
@@ -0,0 +1,102 @@
...
+        outputPath: 'dist/manifest.json',
+        silent: true
+      });
+
+      const categories = discoverCategories(blocksPath);
+      const distIndexPath = 'dist/index.html';
+      if (fs.existsSync(distIndexPath)) {
+        let indexHtml = fs.readFileSync(distIndexPath, 'utf-8');
+        indexHtml = injectBlocksData(indexHtml, { blocks: manifest.blocks, categories });
+        fs.writeFileSync(distIndexPath, indexHtml);
```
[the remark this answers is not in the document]
nickonometry: Fixed this, fetching the JSON now instead of injecting the contents directly into the window object

.github/workflows/deploy-storybook-forge.yml
DRiFTy17, line 1: Rather than modifying the Forge storybook deployment to include blocks, we should just publish the blocks their own directory in the `gh-pages` branch here:
https://github.com/tyler-technologies-oss/forge/tree/gh-pages/docs

Then we can have a separately workflow that just builds and deploys blocks to GitHub Pages. There is only a single GitHub Pages "deployment" in a sense anyway, but keeping these two things separate would be beneficial.

blocks/manifest.json
DRiFTy17, line 1: We may be able to ignore committing this manifest file to source control?

.mcp.json
DRiFTy17, line 1: Looks like this file is empty. Might as well remove it.

CLAUDE.md
```diff
@@ -64,3 +64,27 @@ DO NOT add comments unless code is complex. Defer to configs.
...
+
+- ALL Forge apps typically start with the `forge-app-layout` component. It scaffolds an entire app structure—header, sidenav, content area, and footer—with built-in responsiveness. Use the "application-layout" category blocks as primary references for app structure.
+- When creating new blocks with AI, ALWAYS review existing blocks first (via MCP tools) for references, code patterns, and conventions. This ensures consistency across all blocks.
+- NEVER use placeholder attributes on form fields unless specifically asked to do so.
+- Use `forge-file-picker` component for file uploads, not `<input type="file">`.
+- When using `forge-app-layout` or `forge-app-bar`, the page title is the h1, so content headings should start at h2.
+- `forge-select` uses a `label` attribute for its label, not a slotted `<label>` element (unlike `forge-text-field`).
+- When creating forms, reference blocks from the "forms" and "application-layout" categories for typography, spacing, and layout patterns.
+- When creating cards that require a header, body, and footer, use `forge-structured-card` from `@tylertech/forge-extended` as it provides the necessary layout structure with dedicated slots.
```
DRiFTy17, line 90: Were these testing instructions or do we want this in the root of the repo for a specific reason? I assume maintainers will install the claude plugin at the user level of their machine so this may not be necessary.

.github/workflows/storybook-preview-forge.yml
```diff
@@ -39,6 +39,8 @@ jobs:
 
       - run: pnpm install --frozen-lockfile
 
+      - run: pnpm turbo run build --filter=@tylertech/forge
```
DRiFTy17: Is this needed? The `build:storybook` command _should_ handle building Forge first shouldn't it?

.github/workflows/blocks-preview-cleanup.yml
```diff
@@ -0,0 +1,51 @@
...
+            try {
+              await github.rest.repos.deleteAnEnvironment({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                environment_name: envName
+              });
+              console.log(`Deleted environment: ${envName}`);
+            } catch (e) {
+              console.log(`Environment ${envName} not found or already deleted`);
+            }
```
DRiFTy17, line 57: Since this workflow will run for every PR merge, we should probably add an output boolean to the remove preview step above, and conditionally run these last two steps based on that output value.

CLAUDE.md
nickonometry: Yeah this was put in place for the initial development of blocks, I'll remove it

blocks/manifest.json
nickonometry: Done

.mcp.json
nickonometry: done

.github/workflows/blocks-preview-cleanup.yml
nickonometry: Done

samrichardsontylertech (approved):

PR #1147, opened 2026-05-05
feat(breadcrumb): add breadcrumb

Alex-Oxthorn:
## Summary
Adds the breadcrumb component

## Checklist
- [X] Tests added/updated
- [X] Docs updated (if applicable)
- [X] Changeset added (`pnpm changeset`)


packages/forge/src/lib/breadcrumb/breadcrumb/breadcrumb-component-delegate.ts
DRiFTy17, line 1: The delegate pattern is primarily a feature used by the table component, and while we do have some other components that are implemented as delegates, it's generally a legacy pattern that we don't apply to new components anymore.

Long story short, you can remove this delegate altogether unless you feel strongly that it's needed.

packages/forge/src/lib/breadcrumb/breadcrumb/breadcrumb.ts
DRiFTy17, line 1: We typically refer to this component in the plural as in "breadcrumbs". Others can chime in, but I think we should rename the parent component.

packages/forge/src/lib/breadcrumbs/breadcrumbs-item/breadcrumbs-item.scss
```diff
@@ -0,0 +1,65 @@
...
+    @include state-layer.provide-theme(
+      (
+        color: #{token(text-color)}
+      )
+    );
+  }
+}
+
+forge-focus-indicator {
+  @include focus-indicator.provide-theme(
```
DRiFTy17, line 32: We may want to set the `outward-offset` to `0px` here as well

packages/forge/src/lib/breadcrumb/crumb/crumb.scss
```diff
@@ -0,0 +1,65 @@
...
+      )
+    );
+  }
+}
+
+forge-focus-indicator {
+  @include focus-indicator.provide-theme(
+    (
+      color: #{token(text-color)},
+      shape: 4px
```
DRiFTy17: We should use the `medium` shape token here rather than hard coding a value

packages/forge/src/lib/breadcrumb/crumb/crumb.scss
```diff
@@ -0,0 +1,65 @@
...
+:host([hidden]) {
+  display: none;
+}
+
+.forge-crumb {
+  @include tokens;
+  @include base;
+}
+
+.forge-crumb__link {
```
DRiFTy17: We don't need to use the BEM naming convention within the shadow DOM since it's encapsulated natively. The top-level `forge-crumb` class is fine, but we don't need to prefix with `forge-crumb__*` everything else.

packages/forge/src/dev/pages/breadcrumb/breadcrumb.ts
```diff
@@ -0,0 +1,124 @@
...
+const secondarySwitch = document.getElementById('opt-secondary') as SwitchComponent;
+const siblingsSwitch = document.getElementById('opt-siblings') as SwitchComponent;
+const addBtn = document.getElementById('btn-add-crumb');
+const removeBtn = document.getElementById('btn-remove-crumb');
+
+iconsSwitch.checked = showIcons;
+secondarySwitch.checked = showSecondary;
+siblingsSwitch.checked = showSiblings;
+
+if (showHomeSwitch) {
```
DRiFTy17: You probably don't need these `if` checks around each since we know they are there (and they are typed properly above).

packages/forge/src/lib/breadcrumb/breadcrumb/_core.scss
```diff
@@ -0,0 +1,49 @@
...
+  gap: #{token(gap)};
+  list-style: none;
+  margin: 0;
+  padding: 0;
+  flex-wrap: nowrap;
+}
+
+@mixin separator {
+  color: #{token(separator-theme)};
+  padding-top: 4px;
```
DRiFTy17: We should use spacing tokens for any values (xxsmall in this case)

packages/forge/src/lib/breadcrumb/breadcrumb/breadcrumb-constants.ts
DRiFTy17, line 1: We're trying to move away from the "constants" file concept for net new components. We can export these from the component modules now. See newer components like the key or meter for examples, and we can remove this file.

packages/forge/src/lib/breadcrumb/breadcrumb/breadcrumb.scss
```diff
@@ -0,0 +1,34 @@
...
+:host([hidden]) {
+  display: none;
+}
+
+.forge-breadcrumb {
+  @include tokens;
+  @include base;
+}
+
+.forge-breadcrumb__separator {
```
DRiFTy17: Same with removing BEM in this file as well.

packages/forge/src/lib/breadcrumb/crumb/crumb-constants.ts
DRiFTy17, line 1: This constants file can be removed the same way as the other one.

packages/forge/src/lib/breadcrumb/crumb/_core.scss
```diff
@@ -0,0 +1,69 @@
...
+@mixin base {
+  display: flex;
+  align-items: flex-start;
+}
+
+@mixin link {
+  color: #{token(text-color)};
+  cursor: pointer;
+  padding: 4px 8px;
+  border-radius: 4px;
```
DRiFTy17: More updates for referencing existing tokens for padding, margin, border-radius... etc. here and elsewhere in this file.

packages/forge/src/lib/core/styles/tokens/breadcrumb/_tokens.scss
```diff
@@ -0,0 +1,13 @@
+@use 'sass:map';
+@use '../../spacing';
+@use '../../theme';
+@use '../../utils';
+
+$tokens: (
+  gap: utils.module-val(breadcrumb, gap, 0px),
+  separator-theme: utils.module-val(breadcrumb, separator-theme, theme.variable(text-low))
```
DRiFTy17: Maybe rename this to `separate-color` for consistency with other components?

packages/forge/src/lib/core/styles/tokens/crumb/_tokens.scss
```diff
@@ -0,0 +1,15 @@
+@use 'sass:map';
+@use '../../spacing';
+@use '../../theme';
+@use '../../typography';
+@use '../../utils';
+
+$tokens: (
+  text-color: utils.module-val(crumb, text-color, theme.variable(primary)),
+  icon-theme: utils.module-val(crumb, icon-theme, theme.variable(text-low)),
```
DRiFTy17: Maybe rename this to `icon-color` for consistency with other tokens/components?

packages/forge/src/lib/breadcrumb/breadcrumb/breadcrumb.ts
```diff
@@ -0,0 +1,244 @@
...
+
+  /** @deprecated Used for compatibility with legacy Forge @customElement decorator. */
+  public static [CUSTOM_ELEMENT_DEPENDENCIES_PROPERTY] = [CrumbComponent, IconButtonComponent, IconComponent, MenuComponent];
+
+  /**
+   * The breadcrumb items to render.
+   * @default []
+   */
+  @property({ attribute: false })
+  public crumbs: ICrumbConfiguration[] = [];
```
DRiFTy17: I think having a configuration-based approach for this is good to have, but we should also offer a `<slot>` to allow devs to manually provide their `<forge-crumb>` components declaratively via the HTML. Curious if others agree as well.

packages/forge/src/lib/breadcrumb/breadcrumb/breadcrumb.ts
```diff
@@ -0,0 +1,244 @@
...
+  @property({ type: Boolean, attribute: 'show-home' })
+  public showHome = false;
+
+  /**
+   * The icon name for the separator between crumbs.
+   * @attribute
+   * @default 'slash_forward'
+   */
+  @property()
+  public separator = 'slash_forward';
```
DRiFTy17: Do we want to allow for overriding this? I don't recall if the design mocks allowed for it, do you know? If not, we should just keep this internal until the need comes up to allow for customizing it.

packages/forge/src/lib/breadcrumb/breadcrumb/breadcrumb.ts
```diff
@@ -0,0 +1,244 @@
...
+
+  public override firstUpdated(): void {
+    setDefaultAria(this, this.#internals, { role: 'navigation', ariaLabel: 'Breadcrumb' });
+    this.#setupResizeObserver();
+  }
+
+  public override updated(changedProperties: PropertyValues<this>): void {
+    if (changedProperties.has('crumbs') || changedProperties.has('showHome') || changedProperties.has('separator')) {
+      this._collapsed = false;
+      requestAnimationFrame(() => {
```
DRiFTy17: Instead of using `requestAnimationFrame()` in Lit, it's typically better to use `await this.updateComplete` to ensure that state updates are propagated by Lit after modifying a state property (`_collapsed` in this case).

packages/forge/src/lib/breadcrumb/breadcrumb/breadcrumb.ts
```diff
@@ -0,0 +1,244 @@
...
+      value: i,
+      secondaryLabel: c.secondary,
+      leadingIcon: c.icon,
+      leadingIconType: 'component'
+    }));
+
+    return html`
+      <li class="forge-breadcrumb__crumb-item">
+        <forge-menu .options=${menuOptions} @forge-menu-select=${this.#handleCollapsedMenuSelect}>
+          <forge-icon-button class="forge-breadcrumb__collapsed-trigger" aria-label="Show all breadcrumbs">
```
DRiFTy17: We will need to allow for customizing any internal labels such as this `aria-label` for localization purposes...

packages/forge/src/lib/breadcrumb/crumb/crumb.ts
```diff
@@ -0,0 +1,138 @@
...
+      label: route.label,
+      value: route.path ?? route.label,
+      secondaryLabel: route.secondary,
+      leadingIcon: route.icon,
+      leadingIconType: 'component'
+    }));
+
+    return html`
+      <forge-menu .options=${menuOptions} @forge-menu-select=${this.#handleSiblingSelect} dense>
+        <forge-icon-button class="forge-crumb__sibling-trigger" aria-label="Sibling routes">
```
DRiFTy17: Another instance of internal `aria-label` here for localization hooks.

packages/forge/src/lib/breadcrumb/breadcrumb/breadcrumb.ts
Alex-Oxthorn: Talked about this in private but leaving this here for visibility to others. My team's asking for the ability to override this, as our use case is not strictly a representation of navigation. We're using it to represent a series of events in an audit trail, and our designer requested that the separator be a chevron

packages/forge/src/lib/breadcrumb/crumb/crumb.scss
samrichardsontylertech: [The key item is a good, simple demonstration of this](https://github.com/tyler-technologies-oss/forge/blob/main/packages/forge/src/lib/key/key-item/key-item.scss)

packages/forge/src/lib/breadcrumb/breadcrumb/breadcrumb-constants.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export interface ICrumbConfiguration {
+  label: string;
+  path?: string;
+  icon?: string;
+  secondary?: string;
+  siblingRoutes?: ICrumbConfiguration[];
+}
+
+export interface IBreadcrumbSelectEventData {
```
samrichardsontylertech: I'll let Kieran chime in with his opinions but I've been moving away from custom events with data objects and trying to use more native event types with only data in the spec — consumers can (and maybe should?) get this information from the event target.

packages/forge/src/lib/breadcrumb/breadcrumb/breadcrumb.ts
```diff
@@ -0,0 +1,244 @@
...
+import '../../menu/menu.js';
+import '../crumb/crumb.js';
+
+import { ICrumbConfiguration, IBreadcrumbSelectEventData, BREADCRUMB_CONSTANTS, BREADCRUMB_TAG_NAME } from './breadcrumb-constants.js';
+import { CrumbComponent } from '../crumb/crumb.js';
+import type { IMenuOption } from '../../menu/menu-constants.js';
+
+import styles from './breadcrumb.scss';
+
+export interface IBreadcrumbComponent {
```
samrichardsontylertech: We've been removing interfaces in favor of just using the component class. It has benefits with access to the underlying `HTMLElement`/`LitElement`, etc.

packages/forge/src/lib/breadcrumb/breadcrumb/breadcrumb-constants.ts
DRiFTy17: Yeah this is a fundamental shift in how we have been doing things in the past. I'm with Sam on this in that we should prefer standard `Event` first going forward, since using `target` on an event to access the current state is typical unless there is intermediary state that need to be provided that can't be accessed through the public API.

There is a further discussion here around `CustomEvent` vs `extends Event` (subclassing `Event`) but that's a separate topic we need to iron out.

Given that this is net new, I would try to use standard `Event` if possible, and if we needed we could switch to `CustomEvent` in the future if needed later without breaking changes (although unlikely).

packages/forge/src/lib/breadcrumb/breadcrumb/breadcrumb.ts
```diff
@@ -0,0 +1,244 @@
...
+  @property()
+  public separator = 'slash_forward';
+
+  @state()
+  private _collapsed = false;
+
+  @state()
+  private _expandedContentWidth = 0;
+
+  @query('.forge-breadcrumb')
```
samrichardsontylertech: You might watch to cache this — `@query('.forge-breadcrumb', true)`

packages/forge/src/lib/breadcrumb/breadcrumb/breadcrumb.ts
```diff
@@ -0,0 +1,244 @@
...
+  private _listEl?: HTMLElement;
```
samrichardsontylertech: We know this will always exist so you can give it a non-null assertion, it'll make the rest of the code cleaner

packages/forge/src/lib/breadcrumb/breadcrumb/breadcrumb.ts
```diff
@@ -0,0 +1,244 @@
...
+
+  #internals: ElementInternals;
+  #containerObserver?: ResizeObserver;
```
samrichardsontylertech: You can replace this with the `ForgeResizeObserver`

packages/forge/src/lib/breadcrumb/breadcrumb/breadcrumb.ts
```diff
@@ -0,0 +1,244 @@
...
+  }
+
+  public override disconnectedCallback(): void {
+    super.disconnectedCallback();
+    this.#containerObserver?.disconnect();
+  }
+
+  public render(): TemplateResult {
+    return html`
+      <nav part="root">
```
samrichardsontylertech: Do we want to expose the semantic elements to the light DOM for easier access? This `<nav>` will have to be labelled at the very least. Coordinating that across the shadow boundary isn't very ergonomic.

samrichardsontylertech: This may even be something we want to leave to the consumer. Semantically loading components hasn't worked out well for us in the past.

packages/forge/src/lib/breadcrumb/breadcrumb/breadcrumb.ts
```diff
@@ -0,0 +1,244 @@
...
+        <ol class="forge-breadcrumb">
```
samrichardsontylertech: Same with this list. Our other list-like components put the role on the host.

packages/forge/src/lib/breadcrumb/breadcrumb/breadcrumb.ts
```diff
@@ -0,0 +1,244 @@
...
+    `;
+  }
+
+  #renderHomeButton(): TemplateResult | typeof nothing {
+    if (!this.showHome) {
+      return nothing;
+    }
+    return html`
+      <li class="forge-breadcrumb__crumb-item">
+        <forge-icon-button class="forge-breadcrumb__home-button" aria-label="Home" @click=${this.#handleHomeClick}>
```
samrichardsontylertech: This should probably be a link instead of a button, or be configurable. I'd also give some thought to how this might integrate with frameworks. Do we want to try supporting Angular's router? (That's been a goal in our other components)

packages/forge/src/lib/breadcrumb/breadcrumb/breadcrumb.ts
```diff
@@ -0,0 +1,244 @@
...
+          <forge-icon name="home"></forge-icon>
+        </forge-icon-button>
```
samrichardsontylertech: Should this icon button have a tooltip?

packages/forge/src/lib/breadcrumb/breadcrumb/breadcrumb.ts
```diff
@@ -0,0 +1,244 @@
...
+
+  #renderSeparator(): TemplateResult {
+    return html`<forge-icon class="forge-breadcrumb__separator" .name=${this.separator}></forge-icon>`;
+  }
+
+  #renderExpanded(): TemplateResult[] {
+    return this.crumbs.map((crumb, index) => {
+      const isLast = index === this.crumbs.length - 1;
+      return html`
+        <li class="forge-breadcrumb__crumb-item">
```
samrichardsontylertech: I wonder if we want the crumb itself to be the list item instead, for consistency with our other components and to keep semantics exposed.

packages/forge/src/lib/breadcrumb/breadcrumb/breadcrumb.ts
```diff
@@ -0,0 +1,244 @@
...
+        ? html`
+            <li class="forge-breadcrumb__crumb-item">
+              <forge-crumb .crumb=${lastCrumb} .index=${this.crumbs.length - 1} active> </forge-crumb>
+            </li>
+          `
+        : nothing}
+    `;
+  }
+
+  #setupResizeObserver(): void {
```
samrichardsontylertech: This can more or less be replaced with `ForgeResizeObserver`

packages/forge/src/lib/breadcrumb/crumb/crumb.ts
```diff
@@ -0,0 +1,138 @@
...
+import '../../menu/menu.js';
+import '../../state-layer/state-layer.js';
+
+import { ICrumbConfiguration, IBreadcrumbSelectEventData, BREADCRUMB_CONSTANTS } from '../breadcrumb/breadcrumb-constants.js';
+import { CRUMB_TAG_NAME } from './crumb-constants.js';
+import type { IMenuOption } from '../../menu/menu-constants.js';
+
+import styles from './crumb.scss';
+
+export interface ICrumbComponent {
```
samrichardsontylertech: You can remove this interface too

packages/forge/src/lib/breadcrumb/crumb/crumb-constants.ts
```diff
@@ -0,0 +1,3 @@
+import { COMPONENT_NAME_PREFIX } from '../../constants.js';
+
+export const CRUMB_TAG_NAME: keyof HTMLElementTagNameMap = `${COMPONENT_NAME_PREFIX}crumb`;
```
samrichardsontylertech: I'm not \*opposed\* to calling this `forge-crumb`, but has the name been workshopped? Our other container-collection components typically use a `some-component`, `some-component-item` naming convention.

packages/forge/src/lib/breadcrumb/breadcrumb/breadcrumb.ts
samrichardsontylertech: Actually I see that you did set the host's role to navigation. I'd remove the inner `<nav>` now, it's just going to add a superfluous navigation element to the a11y tree. Though I'd still argue for making the host a list and expecting a nav to be provided externally.

packages/forge/src/lib/breadcrumb/crumb/crumb.ts
```diff
@@ -0,0 +1,138 @@
...
+      <forge-menu .options=${menuOptions} @forge-menu-select=${this.#handleSiblingSelect} dense>
+        <forge-icon-button class="forge-crumb__sibling-trigger" aria-label="Sibling routes">
+          <forge-icon name="arrow_drop_down"></forge-icon>
+        </forge-icon-button>
+      </forge-menu>
+    `;
+  }
+
+  #handleClick(): void {
+    this.dispatchEvent(
```
samrichardsontylertech: I do kind of wonder whether this event is strictly necessary or if there's a cleaner way to handle interaction

packages/forge/src/lib/breadcrumb/breadcrumb/breadcrumb-constants.ts
samrichardsontylertech: Something we could do that might be nice is add a readonly `index` property to the crumb so it would still be easy to get from the target without the custom event

packages/forge/src/lib/breadcrumb/breadcrumb/breadcrumb-component-delegate.ts
Alex-Oxthorn: removed

packages/forge/src/lib/breadcrumb/breadcrumb/breadcrumb.ts
Alex-Oxthorn: renamed

packages/forge/src/lib/breadcrumb/breadcrumb/breadcrumb.ts
Alex-Oxthorn: removed

packages/forge/src/lib/breadcrumb/crumb/crumb.ts
Alex-Oxthorn: removed

packages/forge/src/lib/breadcrumbs/breadcrumbs-item/breadcrumbs-item.scss
Alex-Oxthorn: added

packages/forge/src/lib/breadcrumb/breadcrumb/breadcrumb.ts
samrichardsontylertech: I wonder if we might want to make a slot for the separator like in the tree, so the implementation isn't tied to our icon library?

[I'm thinking the methods here for reference.](https://github.com/tyler-technologies-oss/forge/blob/e13e6b3d83000bb7775683c31d59de65631a9f41/packages/forge/src/lib/tree/tree/tree.ts#L495-L536) It basically detects changes in the slot and clones the node to all the child components. We could even abstract it out to a directive if the logic ends up in more than one component.

packages/forge/src/lib/breadcrumb/breadcrumb/breadcrumb.ts
samrichardsontylertech: I think either allowing a slotted menu or providing a render callback would fix this (same for all the other hardcoded strings)

packages/forge/src/lib/breadcrumb/breadcrumb/breadcrumb.ts
Alex-Oxthorn: The other frameworks was the reason for a button and event over a link. This allows the consumer to tie into the event and handle it however their framework of choice works. Or at least that was my intention. If there is a better way to do that, I am eager to learn that.

packages/forge/src/lib/breadcrumb/crumb/crumb-constants.ts
Alex-Oxthorn: No idea, seemed the logical name at the time. Either way, I've updated the name to reflect the other naming patterns.

packages/forge/src/lib/breadcrumbs/breadcrumbs/breadcrumbs.ts
```diff
@@ -0,0 +1,433 @@
...
+ * @property {string} [homeTooltip='Home'] - The tooltip text for the home button.
+ * @attribute {string} [home-tooltip='Home'] - The tooltip text for the home button.
+ *
+ * @property {string} [expandLabel='Show all breadcrumbs'] - The aria-label for the collapsed breadcrumbs trigger button.
+ * @attribute {string} [expand-label='Show all breadcrumbs'] - The aria-label for the collapsed breadcrumbs trigger button.
+ *
+ * @property {string} [siblingRoutesLabel='Sibling routes'] - The aria-label for the sibling routes trigger button within each crumb.
+ * @attribute {string} [sibling-routes-label='Sibling routes'] - The aria-label for the sibling routes trigger button within each crumb.
+ *
+ * @fires {BreadcrumbsSelectEvent} forge-breadcrumbs-crumb-select - Dispatched when a crumb is clicked.
```
DRiFTy17, line 63: The type here should just be `Event`

packages/forge/src/stories/components/breadcrumbs/Breadcrumbs.stories.ts
```diff
@@ -0,0 +1,126 @@
...
+    ]
+  },
+  { label: 'Components' }
+];
+
+const meta = {
+  title: 'Components/Breadcrumbs',
+  tags: ['new'],
+  render: args => {
+    const el = document.createElement(component) as BreadcrumbsComponent & HTMLElement;
```
DRiFTy17, line 51: I don't think the `as` cast here is necessary? `components` is a `const` and TypeScript is smart enough to pick up the ambient typings. Just want to avoid this to ensure the wrong pattern doesn't make its way around if others look at this example.

packages/forge/src/dev/pages/breadcrumbs/breadcrumbs.ts
```diff
@@ -0,0 +1,125 @@
...
+
+removeBtn.addEventListener('click', () => {
+  if (baseCrumbs.length > 1) {
+    baseCrumbs.splice(baseCrumbs.length - 2, 1);
+    updateAll();
+  }
+});
+
+window.addEventListener('forge-breadcrumbs-crumb-select', evt => {
+  console.log('Crumb selected:', (evt as CustomEvent).detail);
```
DRiFTy17, line 120: Now that `forge-breadcrumbs-crumb-select` is a standard `Event`, there is no `detail` property anymore...

Also, this may present an issue if the event `target` is the `<forge-breadcrumbs>` element rather than the `<forge-breadcrumb-item>` element since we wouldn't be able to access _which_ breadcrumb item was selected. I assume this is why the `index` getter was added? I'm not sure how I feel about that approach... 

We may want to keep this event as a `CustomEvent` and just include the crumb data in the detail? @samrichardsontylertech thoughts?

packages/forge/src/lib/breadcrumbs/breadcrumbs-item/_core.scss
```diff
@@ -0,0 +1,72 @@
...
+
+@mixin host {
+  display: flex;
+  align-items: flex-start;
+  white-space: nowrap;
+}
+
+@mixin base {
+  display: flex;
+  align-items: flex-start;
```
DRiFTy17, line 17: I assume we're using `align-items: flex-start` here to account for secondary text while keeping the alignment of the separators? Might it be better to introduce a grid layout with two rows? This would also remove the need for setting `padding-top` on the separator which seems to cause alignment issues with different separator icons...

packages/forge/src/lib/breadcrumbs/breadcrumbs-item/_core.scss
```diff
@@ -0,0 +1,72 @@
...
+  );
+}
+
+@mixin active {
+  padding: var(--forge-spacing-xxsmall) var(--forge-spacing-xsmall);
+}
+
+@mixin separator {
+  color: #{breadcrumb-tokens.token(separator-color)};
+  padding-top: var(--forge-spacing-xxsmall);
```
DRiFTy17, line 71: I would like to avoid this, see previous comment related to grid layout.

packages/forge/src/lib/breadcrumbs/breadcrumbs-item/breadcrumbs-item.scss
```diff
@@ -0,0 +1,67 @@
...
+.sibling-trigger {
+  @include sibling-trigger;
+}
+
+.active {
+  @include active;
+}
+
+.separator {
+  @include separator;
```
DRiFTy17, line 62: Is the default separator icon font-size a little too large? Maybe we want `1.5rem` or `1.25rem` instead? Seemed a little strong compared to the mocks.

packages/forge/src/lib/core/styles/tokens/breadcrumbs-item/_tokens.scss
```diff
@@ -0,0 +1,15 @@
+@use 'sass:map';
+@use '../../spacing';
+@use '../../theme';
+@use '../../typography';
+@use '../../utils';
+
+$tokens: (
+  text-color: utils.module-val(crumb, text-color, theme.variable(primary)),
+  icon-color: utils.module-val(crumb, icon-color, theme.variable(text-low)),
+  secondary-text-color: utils.module-val(crumb, secondary-text-color, theme.variable(text-low))
```
DRiFTy17, line 10: This should be `text-medium`

packages/forge/src/lib/breadcrumbs/breadcrumbs-item/breadcrumbs-item.ts
```diff
@@ -0,0 +1,144 @@
...
+    if (this.active || !this.crumb.path) {
+      return html`
+        <span class="active" aria-current=${this.active ? 'page' : nothing}>
+          <span class="label-text">${this.crumb.label}</span>
+        </span>
+      `;
+    }
+
+    return html`
+      <button class="link" type="button" @click=${this.#handleClick}>
```
DRiFTy17, line 80: We may want to allow for the items to alternatively be rendered as an anchor tag? While convenient, the issue with that is framework routers need to be able to hook into it which cannot be done unless the `<a>` is exposed via a slot... We do some janky logic for this in the `<forge-list-item>` with a hidden shadow `<a>` but that's not ideal. Just wanted to bring it up since navigation like this may be better suited as an `<a>`. Thoughts @samrichardsontylertech?

packages/forge/src/lib/breadcrumbs/breadcrumbs/breadcrumbs.ts
```diff
@@ -0,0 +1,433 @@
...
+  @property({ attribute: 'separator-icon-name' })
+  public separatorIconName = 'slash_forward';
+
+  /**
+   * The tooltip text for the home button.
+   * @attribute home-tooltip
+   * @default 'Home'
+   */
+  @property({ attribute: 'home-tooltip' })
+  public homeTooltip = 'Home';
```
DRiFTy17, line 109: Since `<forge-tooltip>` is an inline element, we could use a `<slot>` for this instead? We do this in other components as well is why I mention it.

packages/forge/src/lib/breadcrumbs/breadcrumbs/breadcrumbs.ts
```diff
@@ -0,0 +1,433 @@
...
+   * The aria-label for the sibling routes trigger button within each crumb.
+   * @attribute sibling-routes-label
+   * @default 'Sibling routes'
+   */
+  @property({ attribute: 'sibling-routes-label' })
+  public siblingRoutesLabel = 'Sibling routes';
+
+  #selectedIndex = -1;
+
+  public get index(): number {
```
DRiFTy17, line 129: Is `index` clear enough? This should have `JSDoc` but maybe be renamed? If we decide to keep it at least due to the event issue above...

PR #1157, opened 2026-05-29
feat(toolbar): convert to Lit

geraltofrivendell:
## Summary

Convert toolbar to Lit.

## Checklist
- [x] Tests added/updated


packages/forge/src/lib/toolbar/toolbar.ts
```diff
@@ -61,42 +60,49 @@ declare global {
...
-    }
-  }
+  @property({ type: Boolean, reflect: true }) public inverted = false;
 
-  public get inverted(): boolean {
-    return this._inverted;
+  public render(): TemplateResult {
+    const classes = {
+      'forge-toolbar': true,
+      'forge-toolbar--inverted': this.inverted
```
DRiFTy17: We should update the selector to target this class, rather than the reflected attribute to help prepare for how we will not reflect attributes in the future:
https://github.com/tyler-technologies-oss/forge/blob/dd93b3dbc94585470c25cc5b80f57133ca426576/packages/forge/src/lib/toolbar/toolbar.scss#L51

Additionally, we don't need to use BEM naming convention in the shadow DOM, so we can just use the simple class name of `inverted`.

Swap it to this basically:
```scss
.inverted {
  @include inverted;
}
```

packages/forge/src/lib/toolbar/toolbar.ts
```diff
@@ -61,42 +60,49 @@ declare global {
...
+  public static [CUSTOM_ELEMENT_DEPENDENCIES_PROPERTY] = [];
 
-  public attributeChangedCallback(name: string, oldValue: string, newValue: string): void {
-    switch (name) {
-      case TOOLBAR_CONSTANTS.observedAttributes.INVERTED:
-        this.inverted = coerceBoolean(newValue);
-        break;
-    }
-  }
+  @property({ type: Boolean, reflect: true }) public inverted = false;
```
DRiFTy17, line 74: No need to bother right now, but I wanted you to be aware that once we turn off attribute reflection in the future, we will swap to use CSS custom states via `ElementInternals`. This means we would toggle a state called `:inverted` on this element that would allow consumers to target that custom state via CSS.

DRiFTy17 (commented):
This looks really good overall. It's a simple component of course, but the conversion was very clean.

packages/forge/src/lib/toolbar/toolbar.ts
samrichardsontylertech: 👆

packages/forge/src/lib/toolbar/toolbar.ts
samrichardsontylertech: What Kieran said, but it might not even be worth creating a state for `inverted`. It depends whether we can reasonably expect someone to need to select or style  a toolbar based on whether it's inverted.

samrichardsontylertech (changes requested):
Looks good! Just fix the inverted selector and I'll be ready to approve

packages/forge/src/lib/toolbar/toolbar.ts
geraltofrivendell: Yeah that makes sense. Though I agree it seems like an odd case someone would need to select or style based on inverted.

packages/forge/src/lib/toolbar/toolbar.ts
```diff
@@ -59,7 +59,7 @@ export const TOOLBAR_TAG_NAME: keyof HTMLElementTagNameMap = 'forge-toolbar';
  * @slot after-end - The content to place after the end slot.
  *
  * @cssclass forge-toolbar - Apply to the root element _(required)_.
- * @cssclass forge-toolbar--inverted - Inverts the toolbar so the divider is at the top.
+ * @cssclass inverted - Inverts the toolbar so the divider is at the top.
```
DRiFTy17: One minor confusion here, and it's an area where we overload the component a bit to support the documentation output. This should stay `forge-toolbar--inverted` because there is a CSS-only implementation in `forge-toolbar.scss` (which is not used by the component itself). Confused yet?

We use the component JSDoc to document those CSS-only stylesheets via the `@cssclass` custom JSDoc tag for the CEM (Custom Elements Manifest).

Long story short, these `@cssclass` tags only pertain to the stylesheets, not the internal shadow DOM classes. Sorry I didn't mention that earlier.

geraltofrivendell: No worries that makes sense

DRiFTy17 (approved):

samrichardsontylertech (approved):
Looks good 👍

PR #1164, opened 2026-06-10
feat(tabs): add support for named tabs

samrichardsontylertech:
## Summary
Added a `name` property to the Tab component. The active tab can be set or retrieved from the Tab Bar component via the `activeTabName` property.

To support this alongside the index-based `activeTab` property, both `activeTab` and `activeTabName` have been rewritten as getter/setter pairs referring back to `#activeTabElement` as the shared source of truth.

Also rewrote active tab state syncing to stabilize within one update cycle, down from the three-or-so cycles required previously.

## Checklist
- [x] Tests added/updated

## Breaking Changes
None


samrichardsontylertech:
Do we want the new property to be `activeTabName` or `value`? `value` is usually associated with form elements but the `<li>` element also supports it to set its ordinal value, so there's some precedent for flexibility.

packages/forge/src/stories/components/tabs/Tabs.mdx
```diff
@@ -53,6 +54,42 @@ When disabling tabs you should either choose to disable all tabs together throug
...
+
+To get the name of the active tab when the active tab changes, listen for the `forge-tab-bar-change`
+event on the `<forge-tab-bar>` element and access the `activeTabName` property on the event's
+target.
+
+```ts
+const tabBar = document.querySelector('forge-tab-bar');
+tabBar?.addEventListener('forge-tab-bar-change', (event) => {
+  const activeTabName = (event.target as HTMLForgeTabBarElement).activeTabName;
+  console.log('Active tab name:', activeTabName);
```
DRiFTy17: Now that `activeTabName` no longer syncs automatically, these docs are incorrect... Will need to access the new tab via event detail.

packages/forge/src/lib/tabs/tab-panel/tab-panel.ts
```diff
@@ -0,0 +1,346 @@
...
+    }
+
+    const shouldOpen = this.forElement.active;
+    if (shouldOpen === this.open) {
+      return;
+    }
+
+    const beforeToggleEvent = this.#dispatchBeforeToggleEvent(shouldOpen);
+    if (beforeToggleEvent.defaultPrevented) {
+      evt.preventDefault();
```
DRiFTy17: The `forge-tab-request-sync` event is not cancelable, so this isn't doing anything currently. Was this supposed to handle a specific case?

packages/forge/src/stories/components/tabs/Tabs.mdx
samrichardsontylertech: I'll add a name field to the event detail

packages/forge/src/lib/tabs/tab-panel/tab-panel.ts
samrichardsontylertech: Removing `beforetoggle` for now, we can reimplement it later if we figure out how to do it cleanly

DRiFTy17 (approved):

PR #1167, opened 2026-06-15
feat(tabs): create tab panel component

samrichardsontylertech:
## Summary
Added a `<forge-tab-panel>` component that syncs with a tab to automatically manage visibility and semantics.

## Checklist
- [x] Tests added/updated

## Breaking Changes
None


packages/forge/src/lib/tabs/tab-panel/tab-panel.ts
```diff
@@ -0,0 +1,343 @@
...
+import { toggleState } from '../../core/utils/utils.js';
+import { TAB_CONSTANTS } from '../tab/tab-constants.js';
+import type { TabComponent } from '../tab/tab.js';
+import { FocusIndicatorComponent } from '../../focus-indicator/index.js';
+
+import styles from './tab-panel.scss';
+
+export const TAB_PANEL_TAG_NAME: keyof HTMLElementTagNameMap = `${COMPONENT_NAME_PREFIX}tab-panel`;
+
+export type TabPanelFocusMode = 'auto' | 'off';
```
nickonometry: Could just be me but was a little confused by values at quick glance. Do you think it would make more sense if the values were panel | tab, and then:

public focusedElement: TabPanelFocusMode = 'panel';

or

public focusedElement: TabPanelFocusMode = 'tab';

nickonometry (approved):

packages/forge/src/lib/tabs/tab-panel/tab-panel.ts
samrichardsontylertech: Maybe? It doesn't set focus to the tab when it's off though. There might be a better property name we could use.

PR #1181, opened 2026-06-30
feat(date-time): add forge-date-time-picker and forge-date-time-field components

dannydel:
## Summary

Adds two new Lit-based components for combined **date + time** selection, designed to compose like `<input list>` / `<datalist>`:

- **`forge-date-time-picker`** — a composite that pairs `forge-calendar` with time selection (`single`, `range`, or `slots`/booking-style), usable inline (static card) or anchored as a popover/overlay. Form-associated via `ElementInternals`.
- **`forge-date-time-field`** — a form-associated, masked date/time text input that optionally links to a standalone picker by IDREF (`picker`) or element reference (`pickerElement`), syncing value bidirectionally.

Both gain an orthogonal **`date-mode`** (`single` | `range`) axis, enabling full multi-day date+time ranges with a range-select calendar, quick-range presets, duration summary, deferred Apply/Cancel commit (with `auto-commit` opt-out), and end-after-start validation.

## Highlights

- **Two composable components** linked by value only — the field renders a calendar toggle only when a picker is linked; the picker works standalone (inline or anchored).
- **Time modes:** `single`, `range` (same-day start/end), and `slots` (booking-style listbox generated from `min-time`/`max-time`/`step`).
- **Date ranges:** range-select calendar, quick-range presets (Today, This Week, Next 7 Days, This Month), duration summary, deferred commit.
- **Value model:** `Temporal.PlainDateTime` by default, with `iso` and `date` `value-mode` options; date-only `min`/`max`/`value` parse as local wall-clock time (no off-by-one west of UTC).
- **Mobile:** on phones (≤599px) the picker opens as a full-height `forge-bottom-sheet` and a linked field shows a two-line start/end value display; desktop behavior is unchanged.
- **Accessibility (WCAG 2.1 AA):** `role="group"` host named from its label, listbox semantics for slot mode, live-region selection announcements, full keyboard nav (arrows, Home/End, Enter/Space, digit typeahead), `aria-expanded` on the toggle, `role="dialog"` with an accessible name on the anchored card. Also fixes related `forge-calendar` axe violations (month-nav button names, empty grid cells).
- **Theming:** CSS custom properties + Forge design tokens; optional primary-colored `summary` side panel; `header` / `footer-*` / `time-label` slots.

## Breaking / migration notes

- `time-mode="range"` now means a **same-day** time range; it is expressed as `date-mode="single" time-mode="range"`. For a multi-day date range, set `date-mode="range"`.
- `forge-date-time-field` no longer embeds a picker/popover. Link a standalone `forge-date-time-picker` via the `picker` IDREF attribute or `pickerElement` property. The field's `step`, `min-time`, and `max-time` are removed — set them on the linked picker instead.

## Documentation

- Storybook stories + MDX docs pages for both components (single/range/slots modes, presets, summary, orientation, standalone field).
- Dev pages for interactive local testing.

## Changesets

7 changesets included (net **minor**): 4 feature (`minor`) + 3 fix (`patch`), covering the new components, the `date-mode` range axis, small-screen support, the composable split, and correctness/accessibility fixes.

## Testing

- `pnpm build` (library typecheck + bundle) and `pnpm lint` pass locally.
- Storybook builds and both docs pages + story canvases render and were spot-checked in a browser.
- The full browser-based vitest suite runs in CI on a clean checkout.


MikeMatusz:
@DRiFTy17 is there any way to get a Storybook staging deployment for this PR?

DRiFTy17:
> @DRiFTy17 is there any way to get a Storybook staging deployment for this PR?

Unfortunately due to it being a forked branch, the GitHub workflow can't use the permissions from the fork for deployment...

dannydel:
Tomorrow I'll bring it over to the cloned repo from the fork and create a new PR, if you are cool with that? @DRiFTy17

DRiFTy17:
> Tomorrow I'll bring it over to the cloned repo from the fork and create a new PR, if you are cool with that? @DRiFTy17 

Sure, that's fine with me 👍

dannydel:
@DRiFTy17 Just realized why I didn't do this in the first place. I don't have the permissions to create a non forked repo PR. Is there anything you would like me to do further or is it okay as it stands?

packages/forge/src/lib/date-time-field/date-time-field.ts
```diff
@@ -0,0 +1,1118 @@
...
+ * @attribute {('temporal'|'iso'|'date')} [value-mode='temporal']
+ * @attribute {string} [name]
+ * @attribute {string} [label]
+ * @attribute {string} [placeholder]
+ * @attribute {boolean} [disabled=false]
+ * @attribute {boolean} [readonly=false]
+ * @attribute {boolean} [required=false]
+ * @attribute {('both'|'date'|'time')} [required-parts='both']
+ * @attribute {boolean} [open=false]
+ * @attribute {boolean} [persistent=false]
```
MikeMatusz: Could use descriptions on these and the properties.

packages/forge/src/lib/date-time-field/date-time-field.ts
```diff
@@ -0,0 +1,1118 @@
...
+  public static formAssociated = true;
+  /** @deprecated */
+  public static [CUSTOM_ELEMENT_NAME_PROPERTY] = DATE_TIME_FIELD_TAG_NAME;
+
+  @property({ attribute: 'date-mode', reflect: true }) public dateMode: DateTimeFieldDateMode = 'single';
+  @property({ attribute: 'time-mode', reflect: true }) public timeMode: TimeMode = 'single';
+  @property({ attribute: 'value-mode', reflect: true }) public valueMode: DateTimePickerValueMode = 'temporal';
+  @property({ reflect: true }) public name = '';
+  @property() public label = '';
+  @property() public placeholder = '';
```
MikeMatusz, line 166: We don't often use the placeholder on date fields, but I wasn't able to get this to show in any of the demos.  Similarly, and perhaps related, I don't ever see the label unfloated, even when the field doesn't have a value and isn't focused.

packages/forge/src/lib/date-time-field/date-time-field.ts
```diff
@@ -0,0 +1,1118 @@
...
+                slot="end"
+                part="toggle"
+                type="button"
+                aria-label="Open date and time picker"
+                aria-haspopup="dialog"
+                aria-expanded=${this._open ? 'true' : 'false'}
+                ?disabled=${this.disabled}
+                @click=${this.#togglePicker}>
+                <forge-icon name="insert_invitation"></forge-icon>
+              </forge-icon-button>
```
MikeMatusz, line 550: This doesn't behave like the picker buttons in the other fields.  It's a separate tab stop here, and the down arrow doesn't open it.

packages/forge/src/lib/date-time-field/date-time-field.ts
```diff
@@ -0,0 +1,1118 @@
...
+      if (key === 'd') {
+        event.preventDefault();
+        this.#applyToday(event.target as HTMLInputElement);
+        return;
+      }
+      if (key === 't') {
+        event.preventDefault();
+        this.#applyCurrentTime(event.target as HTMLInputElement);
+        return;
+      }
```
MikeMatusz, line 788: We may want to consider standardizing these across the date, time, and date-time pickers.  Currently, date supports `n` or `t` for the current date, time only supports `n`.  Having `t` mean "Today" in one context and "Time" here might be confusing.  I do like being able to universally hit `n` for Now, though.

packages/forge/src/lib/date-time-field/date-time-field.ts
```diff
@@ -0,0 +1,1118 @@
...
+      if (target === this._fromInput || target === this._toInput) {
+        this.#setMaskValue(target, timeStr);
+      }
+    } else {
+      this.#setMaskValue(this._timeInput, timeStr);
+    }
+    this.#onTypedInput();
+  }
+
+  #onTypedInput = (): void => {
```
MikeMatusz: Might it make sense to auto-advance to time after date is entered, or be able to arrow/backspace from time into the date field?  Maybe the extra complexity isn't worth it, and we're already used to them being completely different fields right now, but being displayed together, I found myself always trying to keep typing and expecting it to work.

packages/forge/src/stories/components/date-time-field/DateTimeField.mdx
```diff
@@ -0,0 +1,62 @@
...
+time_.
+
+<Canvas of={DateTimeFieldStories.Demo} />
+
+## Date Range
+
+Set `date-mode="range"` to capture a start and end date in two masked inputs. The field validates that the end date is on or after the
+start date and reports an error otherwise.
+
+<Canvas of={DateTimeFieldStories.DateRange} />
```
MikeMatusz, line 22: I can't think of any scenario where I'd want a date-range but time-mode="single", so I wonder if it might be more useful to have this example with both set to range?  I can definitely configure the Demo story that way easily enough, but seems like a more common use case than this.

packages/forge/src/lib/date-time-field/date-time-field.ts
```diff
@@ -0,0 +1,1118 @@
...
+    }
+  }
+
+  public override render(): TemplateResult {
+    // On phones, a linked field becomes a tappable two-line value display that
+    // opens the picker's bottom sheet; masked typing stays available on desktop
+    // and on unlinked fields (which have no sheet to open).
+    if (this._isPhone && this._pickerLinked) {
+      return this.#renderMobile();
+    }
```
MikeMatusz, line 513: Would it be possible to trigger this base on CSS (container query or just allow wrapping with min-widths) rather than an isPhone property?  With dynamic rendering, or if this were to appear in a dialog/sidebar, it might be nice to use the mobile layout if space is constrained, especially with both date and time modes set to range, since right now that requires a width of ~450px to not have display issues, which is pretty wide.

packages/forge/src/lib/date-time-field/date-time-field.ts
```diff
@@ -0,0 +1,1118 @@
...
+  #renderInputs(): TemplateResult {
+    if (this.timeMode === 'slots') {
+      return this.#renderSlotsMasked();
+    }
+    const range = this.#isRangeValue();
+    const dateRange = this.dateMode === 'range';
+    const timeRange = this.timeMode === 'range';
+    // Inputs must be top-level default-slot children so forge-text-field detects
+    // them (assignedElements is not deep). Grouping is done with margins in SCSS.
+    return html`
```
MikeMatusz, line 1072: Not sure where to drop this comment, but the spacing between the inputs sometimes seems a little strange, e.g. the following screenshot, where the gap between the second date and time is larger than the first, and the gap before the hyphen is very large:

Image

packages/forge/src/lib/date-time-field/date-time-field.scss
```diff
@@ -0,0 +1,205 @@
...
+  background: transparent;
+}
+
+// Empty/incomplete masked segments render their guide (e.g. __/__/____) in a
+// muted color so they read like a placeholder rather than stark entered text.
+// `input.` keeps this winning over the [part] color rules regardless of order.
+input.guide-muted {
+  color: var(--forge-theme-text-low, GrayText);
+}
+
```
MikeMatusz, line 80: This felt a little weird, and I'm not sure had sufficient contrast in dark theme.  I also saw at least twice where the date remained muted even after being complete.

I do like the concept, but maybe we need an intermediate color, stronger than the empty placeholder, but not quite as strong as when it's complete.

MikeMatusz (changes requested):
I pulled the branch and ran it locally, and primarily focused on the field variant of the component.  I still need to go through the code in greater depth, but wanted to get my functional observations to you sooner since my capacity to review is limited right now.  I'm really excited to see this component coming together!

packages/forge/src/lib/date-time-field/date-time-field.ts
dannydel: I like that idea. It does feel more intuitive to just keep typing or backspacing compared to tabbing. I'll look into this.

packages/forge/src/lib/date-time-field/date-time-field.ts
dannydel: Yeah added 'n' just out of constantly wanting to tap it due to what I use it for in obsidian. I can remove for now and we can add an item to think about it later or something. Would you like these standardized to a over arching structure they all inherit from?

MikeMatusz: I vote to keep `n` for sure, since that works consistently in all three pickers.  My concern was more around `t` having a different meaning here compared to the forge-time-picker, though `d` and `t` for date and time are logical in context.

packages/forge/src/lib/date-time-field/date-time-field.ts
dannydel: Fixed — added `- description` to every existing `@attribute` entry, filled in the missing ones (`date-mode`, `locale`, `use-24-hour-time`, `allow-seconds`, `min`, `max`, `popover-placement`), and added `@property` entries for the non-attribute members (`value`, `pickerElement`, `form`, `labels`, `validity`, `validationMessage`).

packages/forge/src/lib/date-time-field/date-time-field.ts
dannydel: Fixed!

packages/forge/src/lib/date-time-field/date-time-field.ts
dannydel: I kept it the same for now, but wanted to keep this unresolved just to get your actual input on direction. Keep `n`(now)  and `d` (today) and add in current time to it; then drop `t` entirely?

packages/forge/src/lib/date-time-field/date-time-field.ts
dannydel: Implemented — completing a date segment auto-advances focus into its paired time segment, and ArrowLeft/ArrowRight/Backspace step between a date segment and its paired time segment at the boundary (still respects Ctrl/Cmd/Alt/Shift so native word-jump and selection shortcuts keep working).

packages/forge/src/stories/components/date-time-field/DateTimeField.mdx
dannydel: Yeah, I agree updated both with range.

packages/forge/src/lib/date-time-field/date-time-field.ts
dannydel: Fixed the concrete inconsistency: the range-separator margin and the inter-segment margin were two different hardcoded values (0.4em vs 0.3em) — unified into one shared spacing value so the rhythm is consistent.

packages/forge/src/lib/date-time-field/date-time-field.scss
dannydel: Added a third tone in between — empty/partial/complete are now visually distinct (partial reuses the existing `--forge-theme-text-medium` token, so no new one-off color). Also fixed a real bug behind "stayed muted after being complete": the tone was only recalculated on blur/Enter, so it could go stale while actively typing — now updates live on every keystroke.

packages/forge/src/lib/date-time-field/date-time-field.ts
dannydel: This one I'll need to look further into. A container query should be able to handle this.

packages/forge/src/lib/date-time-field/date-time-field.ts
dannydel: Went a step further — added a one-line doc comment directly above each `@property()` declaration itself, on top of the class-level summary block, so the properties have descriptions in both places.

packages/forge/src/lib/date-time-field/date-time-field.ts
dannydel: What would be most ideal here? I have an idea but want to make sure I have clarity.

dannydel: So most recent round of commits I updated so that it does the following:
- Follows inset label behavior unless (show-mask is set to on and persist-mask is set to on)
- Placeholder does the same and will show if another variation of the label is set.

packages/forge/src/lib/date-time-field/date-time-field.ts
dannydel: Removed the _isphone logic and made it query based. It only wraps in range mode and it has the following logic:
- :host { width: max-content; max-width: 100% } — field hugs content, but clamps to its container.
  - .inputs { display: flex; flex-wrap: wrap } — the input row can wrap.
  - Wrap unit = .endpoint (date+time glued, flex: 0 0 auto); separator lives inside the end endpoint. So the only break point is between start and end endpoints.
  
  Demo shows this now.

packages/forge/src/lib/date-time-field/date-time-field.ts
MikeMatusz: I tried comparing to date-picker, and Storybook doesn't even expose a placeholder there.  With the description now, I can see this was only intended for the `phone` configuration, so I think it's fine not to see the placeholder in desktop mode.

The second part, regarding floating the label, is a departure from date/time picker, though.  Those will conditionally float the label and show the mask only when the field has focus or a value.  This field component always floats the label and shows the mask, which will make it inconsistent on a form with other components.  I know Forge was/is considering moving away from the floating style in favor of label-above, but haven't gotten an update on that lately so not sure how far off a decision on that is, but in the current version, it would ideally behave similarly.

packages/forge/src/stories/components/date-time-field/DateTimeField.stories.ts
```diff
@@ -0,0 +1,147 @@
...
+
+type Story = StoryObj;
+
+export const Demo: Story = {};
+
+export const DateRange: Story = {
+  ...standaloneStoryParams,
+  render: () => html`
+    <forge-date-time-field
+      style="width: 360px"
```
MikeMatusz: Might want to bump this to ~450px, it's currently cut off at this width:

Image

packages/forge/src/lib/date-time-field/date-time-field.ts
```diff
@@ -0,0 +1,1362 @@
...
+        shape=${this.shape}
+        theme=${this.theme}
+        ?required=${this.required}
+        ?invalid=${this._invalid}
+        @mousedown=${this.#onFieldPointerDown}>
+        ${this.label ? html`<label slot="label">${this.label}</label>` : nothing} ${this.#renderInputs()}
+        ${this._pickerLinked
+          ? html`
+              <!-- Matches forge-date-picker's toggle: never a separate tab stop, opened via ArrowDown from the input instead. -->
+              <forge-icon-button
```
MikeMatusz, line 538: I think this might also need `density="medium"` to be consistent with date-picker.

packages/forge/src/stories/components/date-time-field/DateTimeField.stories.ts
```diff
@@ -0,0 +1,147 @@
...
+const changeAction = (evt: CustomEvent): void => action('forge-date-time-field-change')(evt.detail);
+const openAction = (evt: CustomEvent): void => action('forge-date-time-field-open')(evt.detail);
+const closeAction = (evt: CustomEvent): void => action('forge-date-time-field-close')(evt.detail);
+
+const meta = {
+  title: 'Components/Date Time Field',
+  render: args => html`
+    <forge-date-time-field
+      style="width: 320px"
+      picker="dtf-demo-picker"
```
MikeMatusz: Might be more of a storybook rendering issue, but if I switch from Docs to Demo and back for this component, it seems like the picker association isn't working, and it just always displays the picker to the side of the field instead of in the dropdown.  Might also be some kind of timing issue with how the `picker` linking works.  Should the picker be slotted instead?  I don't have a preference, just a potential issue I think we should follow up on.

packages/forge/src/lib/date-time-field/date-time-field.ts
```diff
@@ -0,0 +1,1362 @@
...
+ * @attribute {boolean} [persist-mask=false] - Forces the guide to always show when `show-mask` is on, even if a `placeholder` is set (the placeholder no longer overrides).
+ * @attribute {Date|string} [min] - Minimum selectable date/time; forwarded to the linked picker and used for validation.
+ * @attribute {Date|string} [max] - Maximum selectable date/time; forwarded to the linked picker and used for validation.
+ * @attribute {string} [popover-placement] - Placement of the linked picker's popover relative to the field.
+ *
+ * @property {IDateTimePickerComponent | null} [pickerElement=null] - Direct element reference to the linked picker; an alternative to the `picker` IDREF attribute.
+ * @property {DateTimePickerPublicValue} [value] - The current value, shaped by `value-mode`; `null` when empty.
+ * @property {HTMLFormElement | null} form - The form this field participates in (read-only).
+ * @property {NodeList} labels - The labels associated with this field (read-only).
+ * @property {ValidityState} validity - The field's current validity state (read-only).
```
MikeMatusz, line 138: Is this fully implemented? I would expect it not be valid with a partial value entered (e.g. missing date or time part), or an invalid date like `19/47/2026` but it always seems to report that it's valid.

packages/forge/src/lib/date-time-field/date-time-field.ts
```diff
@@ -0,0 +1,1362 @@
...
+        autocomplete="off"
+        placeholder=${this.#segmentPlaceholder('date', isFirst)}
+        aria-label=${ariaLabel}
+        aria-required=${ifDefined(dateRequired ? 'true' : undefined)}
+        aria-invalid=${ifDefined(this._invalid && dateRequired && !hasSegment ? 'true' : undefined)}
+        aria-haspopup=${ifDefined(this._pickerLinked ? 'dialog' : undefined)}
+        aria-expanded=${ifDefined(this._pickerLinked ? (this._open ? 'true' : 'false') : undefined)}
+        ?disabled=${this.disabled}
+        ?readonly=${this.readonly}
+        @blur=${() => this.#onTypedInput()}
```
MikeMatusz, line 1147: Not sure how they accomplish it, but seems like the date- and time-pickers do some kind of coercion on blur, which sometimes I have concerns about, but it's also nice to be able to type a six digit year (e.g. 07/09/26) and have it infer the century digits.  Or for time, to type A or P and have it complete the M.  There doesn't seem to be anything of the sort with this component, and it's required to type all characters manually.  Are we intentionally omitting that?  I need to look closer to see how much code this component was able to share with those.

packages/forge/src/lib/date-time-field/date-time-field.ts
MikeMatusz: Kind of hoping someone else chimes in.  The `n` behavior is definitely consistent and I want to keep it as-is.

My concern is that on the date-picker `t` means today, but here `t` means time, and `d` means today.  Maybe that's fine if consider these completely different contexts, but with how closely related the date-picker and date-time-field are, I worry there could be some confusion where users use one when they mean the other.  Just speculation, though, so I think you should probably leave them as they are unless someone else can provide more guidance.

dannydel: Yeah, your concerns make total sense. I'll just swap `t` to be **today** and remove `d` since it doesn't match other inputs.

packages/forge/src/stories/components/date-time-field/DateTimeField.stories.ts
dannydel: That sounds like an odd race condition with rendering. Looking now!

packages/forge/src/lib/date-time-field/date-time-field.ts
dannydel: I wasn't aware of those coercion helpers so let me look more into this so that I can cover it properly!

packages/forge/src/lib/date-time-field/date-time-field.ts
dannydel: Re-reviewing this as well!

packages/forge/src/lib/date-time-field/date-time-field.ts
MikeMatusz: Looks great, I even through a `resize:true` on the field and put it through the paces and it behaved perfectly, though now that you also fixed the inset label behavior, if the field is empty/unfocused and wraps, the hyphen remains visible:
image

packages/forge/src/lib/date-time-field/date-time-field.ts
dannydel: added in latest commits!

packages/forge/src/lib/date-time-field/date-time-field.ts
dannydel: Fixed in latest commit!

packages/forge/src/stories/components/date-time-field/DateTimeField.stories.ts
dannydel: Found out the issue. The linking logic runs once and when you swap back and forth between demo and docs it must be not anchoring anymore and just not calling the logic to anchor and relink. Fixing that now

packages/forge/src/lib/date-time-field/date-time-field.ts
dannydel: Dang, I thought I fixed the hyphen including on wraps when inset label is active/ placeholder. I'll re-review and also make sure we have a test to take guard against it coming back.

packages/forge/src/lib/date-time-field/date-time-field.ts
dannydel: Resolved in latest round of commits.

packages/forge/src/lib/date-time-field/date-time-field.ts
dannydel: Resolved in the latest round of commits!

packages/forge/src/lib/date-time-field/date-time-field.ts
dannydel: Resolved with latest round of commits. `t` is back to `Today`, kept `n` as a convenience and removed `d` since it doesn't follow our current standard.  Maybe we should pull these standards out into a separate constants for date and time related inputs?

packages/forge/src/lib/date-time-field/date-time-field.ts
dannydel: I can re review on Monday to ensure I am following the standards expected. Feel free to pull the latest and let me know if there are other changes like this needing to be made!

packages/forge/src/lib/date-time-field/date-time-field.ts
```diff
@@ -0,0 +1,1437 @@
...
+  };
+
+  #onFocusIn = (): void => {
+    if (this._focused) {
+      return;
+    }
+    this._focused = true;
+    // Reveal the guide (and float the inset label) immediately, not on the next render, so the
+    // format guide is present the moment the field gains focus.
+    this.#applyMaskGuide();
```
MikeMatusz, line 398: Not sure what's happening, but when I tab into the field, if it doesn't already have a value, the cursor is at the end of the date mask, so I can't just start typing the date.

dannydel: fixed with latest commit. Mask was causing the issue.

packages/forge/src/lib/calendar/calendar-dom-utils.ts
```diff
@@ -186,6 +190,8 @@ export function getHeader(): HTMLElement {
   previousButton.setAttribute('part', CALENDAR_CONSTANTS.parts.PREVIOUS_BUTTON);
   previousButton.id = CALENDAR_CONSTANTS.ids.PREVIOUS_BUTTON;
   previousButton.type = 'button';
+  // The visual tooltip is aria-hidden, so name the button explicitly for assistive tech.
+  previousButton.setAttribute('aria-label', 'Previous');
```
samrichardsontylertech: This change isn't necessary, the tooltip still labels the button even with `aria-hidden`

packages/forge/src/lib/calendar/calendar-dom-utils.ts
```diff
@@ -200,6 +206,8 @@ export function getHeader(): HTMLElement {
   nextButton.setAttribute('part', CALENDAR_CONSTANTS.parts.NEXT_BUTTON);
   nextButton.id = CALENDAR_CONSTANTS.ids.NEXT_BUTTON;
   nextButton.type = 'button';
+  // The visual tooltip is aria-hidden, so name the button explicitly for assistive tech.
+  nextButton.setAttribute('aria-label', 'Next');
```
samrichardsontylertech: Same here, I'd revert both `aria-label` changes

packages/forge/src/lib/date-time-field/date-time-field.scss
```diff
@@ -0,0 +1,148 @@
+// Single source for the rhythm between adjacent segments (date→time within an
+// endpoint, and around the range separator) so the gaps read consistently.
+$segment-gap: 0.3em;
```
samrichardsontylertech: Is this worth making into a token instead of a local variable?

packages/forge/src/lib/date-time-field/date-time-field.scss
```diff
@@ -0,0 +1,148 @@
...
+  background: transparent;
+  // Fits `hh:mm aa` (12h) / `hh:mm` (24h); widened below when seconds are shown.
+  width: 4.75em;
+  text-align: start;
+  // Small gap from the preceding date input within an endpoint.
+  margin-inline-start: $segment-gap;
+}
+
+// Seconds add `:ss`, so reserve more room to keep `hh:mm:ss aa` from clipping.
+:host([allow-seconds]) {
```
samrichardsontylertech: We're moving away from reflected attributes across the library. Instead of using a host selector I'd apply a class to a shadow element and select that (`.allow-seconds { ... }`)

packages/forge/src/lib/date-time-field/date-time-field.scss
```diff
@@ -0,0 +1,148 @@
...
+    width: 6em;
+  }
+}
+
+[part='slot-display'] {
+  cursor: pointer;
+}
+
+:host([readonly]) [part='slot-display'],
+:host([disabled]) [part='slot-display'] {
```
samrichardsontylertech: Same with these two selectors

packages/forge/src/lib/date-time-field/index.ts
```diff
@@ -0,0 +1,20 @@
...
+export * from './date-time-field.js';
+export * from './date-time-field-constants.js';
+export * from './date-time-field-component-delegate.js';
+
+export function defineDateTimeFieldComponent(): void {
+  defineFieldComponent();
+  defineIconComponent();
+  defineIconButtonComponent();
+  defineDateTimePickerComponent();
+  IconRegistry.define(tylIconInsertInvitation);
```
samrichardsontylertech: This would be better placed in a static initialization block in the component that uses it. We have a few examples in other components (open icon, secret, etc.)

packages/forge/src/lib/date-time-picker/_core.scss
```diff
@@ -0,0 +1,184 @@
...
+
+@mixin slot-disabled {
+  background: #{token(slot-disabled-background)};
+  color: #{token(slot-disabled-color)};
+  border-color: #{token(slot-disabled-outline-color)};
+  cursor: not-allowed;
+  text-decoration: line-through;
+}
+
+@mixin live-region {
```
samrichardsontylertech: We have a live announcer service now, you might want to use that instead of rolling your own live region

packages/forge/src/lib/date-time-field/date-time-field.ts
```diff
@@ -0,0 +1,1437 @@
...
+    return untouched ? 'guide-muted' : 'guide-partial';
+  }
+
+  #renderDateInput(part: 'date-input' | 'to-date-input', ariaLabel: string, hasSegment: boolean): TemplateResult {
+    const dateRequired = this.required && this.requiredParts !== 'time';
+    const inputEl = part === 'date-input' ? this._dateInput : this._toDateInput;
+    const isFirst = part === 'date-input';
+    const classes = `${this.#segmentClass(hasSegment, inputEl)}${isFirst && this.#showsFieldPlaceholder() ? ' field-placeholder' : ''}`;
+    return html`
+      <input
```
samrichardsontylertech, line 1135: I'm a little torn on this... our other fields all expect an input to be slotted in by the consumer so they have access to its accessibility attributes and any third party libraries that require direct access to an input still work. On the other hand we *really* tried to move inputs into the shadow DOM during v3 development. I'd like to get @DRiFTy17 's thoughts on this.

packages/forge/src/lib/date-time-picker/date-time-picker.ts
```diff
@@ -0,0 +1,1362 @@
...
+      <div
+        part="root"
+        class=${classMap({ 'forge-date-time-picker': true })}
+        role=${ifDefined(overlayMode ? 'dialog' : undefined)}
+        aria-label=${ifDefined(overlayMode ? 'Date and time picker' : undefined)}
+        data-mode=${this.timeMode}
+        data-orientation=${resolvedOrientation}
+        data-presentation=${sheet ? 'sheet' : 'popover'}>
+        ${this.summary ? this.#renderSummary() : nothing} ${this.summary ? html`<div class="content">${content}</div>` : content}
+        <div part="live-region" class="live-region" role="status" aria-live="polite" aria-atomic="true"></div>
```
samrichardsontylertech: No need for this, you can use `LiveAnnouncer` from Forge Core

packages/forge/src/lib/date-time-picker/date-time-picker.ts
```diff
@@ -0,0 +1,1362 @@
...
+    const content = html`${this.#renderHeader()} ${this.#renderBody(resolvedOrientation)} ${this.#renderFooter()}`;
+    return html`
+      <div
+        part="root"
+        class=${classMap({ 'forge-date-time-picker': true })}
+        role=${ifDefined(overlayMode ? 'dialog' : undefined)}
+        aria-label=${ifDefined(overlayMode ? 'Date and time picker' : undefined)}
+        data-mode=${this.timeMode}
+        data-orientation=${resolvedOrientation}
+        data-presentation=${sheet ? 'sheet' : 'popover'}>
```
samrichardsontylertech: You can just use classes for these
```ts
const classes = classMap({
  'forge-date-time-picker': true,
  [this.timeMode]: true,
  [this.resolvedOrientation]: true,
  sheet,
  popover: !sheet
});
```

packages/forge/src/lib/date-time-picker/_core.scss
```diff
@@ -0,0 +1,184 @@
...
+  color: #{token(color)};
+
+  border-color: #{token(outline-color)};
+  border-width: #{token(outline-width)};
+  border-style: #{token(outline-style)};
+  border-radius: #{token(shape)};
+
+  box-shadow: #{token(elevation)};
+  box-sizing: border-box;
+  padding: #{token(padding)};
```
samrichardsontylertech: I wouldn't bake any border, shadow, or padding styles into the picker's root element. It can be combined with a card component if a design calls for that.

packages/forge/src/lib/date-time-picker/_core.scss
```diff
@@ -0,0 +1,184 @@
...
+  max-width: 100%;
+  max-height: #{token(slot-list-max-height)};
+  overflow-y: auto;
+  padding: 0;
+  margin: 0;
+  list-style: none;
+
+  // Custom scrollbar styling — defer to system defaults but reserve space so the
+  // layout doesn't shift between scrollable and non-scrollable.
+  scrollbar-gutter: stable;
```
samrichardsontylertech: This should use the Forge scrollbar styles instead of the system default

packages/forge/src/lib/date-time-picker/_core.scss
```diff
@@ -0,0 +1,184 @@
...
+}
+
+@mixin summary {
+  display: flex;
+  flex-direction: column;
+  justify-content: flex-start;
+  gap: #{token(section-gap)};
+  background: #{token(summary-background)};
+  color: #{token(summary-color)};
+  min-width: #{token(summary-min-width)};
```
samrichardsontylertech: The summary changes width as you select different dates, which is fairly disorienting. Maybe it should have a static width?

packages/forge/src/lib/date-time-picker/_core.scss
```diff
@@ -0,0 +1,184 @@
...
+
+@forward './token-utils';
+
+@mixin host {
+  display: inline-block;
+  width: fit-content;
+  max-width: 100%;
+  justify-self: start;
+  color: #{token(color)};
+  font-family: inherit;
```
samrichardsontylertech: Here and across all the text in the component it would be better/more consistent to use the standard typography styles
```scss
@use '../core/styles/typography';

@mixin some-mixin {
  @include typography.style(display5);
}
```

packages/forge/src/lib/date-time-picker/date-time-picker.scss
```diff
@@ -0,0 +1,398 @@
...
+.forge-date-time-picker {
+  @include tokens;
+  @include base;
+}
+
+:host([summary]) .forge-date-time-picker {
+  @include base-with-summary;
+  grid-template-columns: auto auto;
+  align-items: stretch;
+  gap: 0;
```
samrichardsontylertech: Any reason these styles aren't included in the mixin?

packages/forge/src/lib/date-time-picker/date-time-picker.scss
```diff
@@ -0,0 +1,398 @@
...
+  @include body;
+}
+
+.body[data-orientation='horizontal'] {
+  @include body-horizontal;
+}
+
+.body[data-orientation='vertical'] {
+  @include body-vertical;
+  gap: 0;
```
samrichardsontylertech: Same re: this style and the mixin. I see that the `gap: 0;`  is also added in the other place `body-vertical` is used.

packages/forge/src/lib/date-time-picker/date-time-picker.scss
```diff
@@ -0,0 +1,398 @@
...
+}
+
+forge-calendar {
+  width: 100%;
+}
+
+.header,
+.footer,
+.time-label {
+  display: contents;
```
samrichardsontylertech: You can probably just remove these elements altogether and keep the bare `<slot>`s since they don't render boxes anyway

packages/forge/src/lib/date-time-picker/date-time-picker.scss
```diff
@@ -0,0 +1,398 @@
...
+}
+
+.header[data-empty='true'],
+.footer[data-empty='true'],
+.time-label[data-empty='true'] {
+  display: none;
```
samrichardsontylertech: We have a `HideWhenEmptyDirective` in `core/utils/lit-utils.ts` that handles detecting and hiding elements without any slotted content for you

packages/forge/src/lib/date-time-picker/date-time-picker.scss
```diff
@@ -0,0 +1,398 @@
...
+}
+
+.footer {
```
samrichardsontylertech, line 96: You should probably create mixins for this and the other selectors without them in this file

packages/forge/src/lib/date-time-picker/date-time-picker.scss
```diff
@@ -0,0 +1,398 @@
...
+  border-top: solid #{token(outline-width)} #{token(divider-color)};
+  padding-top: #{token(gap)};
+  margin-top: #{token(gap)};
+  display: grid;
+  grid-template-columns: 1fr auto 1fr;
+  align-items: center;
+  gap: #{token(gap)};
+}
+
+.footer[data-empty='true'] {
```
samrichardsontylertech: Same note here about `HideWhenEmptyDirective`, but the style is duplicated here anyway so you can just remove it

packages/forge/src/lib/date-time-picker/date-time-picker.scss
```diff
@@ -0,0 +1,398 @@
...
+  flex-direction: column;
+  gap: 2px;
+  min-width: 140px;
+  padding-inline-end: #{token(body-gap)};
+  margin-inline-end: #{token(body-gap)};
+  border-inline-end: solid #{token(outline-width)} #{token(divider-color)};
+  align-self: stretch;
+}
+
+[part='preset'] {
```
samrichardsontylertech, line 196: I think it'd be cleaner to use class selectors for things like this? I might've done it like this in a few old components but I should really go back and fix stuff like that

packages/forge/src/lib/date-time-picker/date-time-picker.scss
```diff
@@ -0,0 +1,398 @@
...
+  padding-block: #{token(slot-padding-block)};
+  padding-inline: #{token(slot-padding-inline)};
+  border: none;
+  border-radius: #{token(shape)};
+  cursor: pointer;
+  user-select: none;
+  transition: background-color 120ms ease;
+
+  &:hover {
+    background: #{token(slot-hover-background)};
```
samrichardsontylertech: This should use the state layer component, but actually it should probably just be a Forge button component for consistency and so we don't have to reimplement basic stuff

packages/forge/src/lib/date-time-picker/date-time-picker.scss
```diff
@@ -0,0 +1,398 @@
...
+  }
+
+  &:focus-visible {
```
samrichardsontylertech: Same thing here, the focus indicator component should handle this but it's already built into the button anyway

packages/forge/src/lib/date-time-picker/date-time-picker.scss
```diff
@@ -0,0 +1,398 @@
...
+  text-overflow: ellipsis;
+  white-space: nowrap;
+}
+
+:host([disabled]) {
+  pointer-events: none;
+}
+
+:host([disabled]) .forge-date-time-picker {
+  opacity: 0.55;
```
samrichardsontylertech: This seems like a none-standard opacity? I need to check what our other components do

packages/forge/src/lib/date-time-picker/date-time-picker.scss
```diff
@@ -0,0 +1,398 @@
...
+  font: inherit;
+  font-size: 0.875rem;
+  text-align: start;
+  padding-block: #{token(slot-padding-block)};
+  padding-inline: #{token(slot-padding-inline)};
+  border: none;
+  border-radius: #{token(shape)};
+  cursor: pointer;
+  user-select: none;
+  transition: background-color 120ms ease;
```
samrichardsontylertech: We have tokens for animations, no need to manually write durations or easing functions

packages/forge/src/lib/date-time-picker/date-time-picker.ts
```diff
@@ -0,0 +1,1362 @@
...
+ * @attribute {string} [from-label='Start time'] - Label for the start-time input in `range` mode.
+ * @attribute {string} [to-label='End time'] - Label for the end-time input in `range` mode.
+ *
+ * @cssproperty --forge-date-time-picker-summary-background - Summary panel background color.
+ * @cssproperty --forge-date-time-picker-summary-color - Summary panel text color.
+ * @cssproperty --forge-date-time-picker-summary-min-width - Summary panel min-width.
+ * @cssproperty --forge-date-time-picker-summary-padding - Summary panel padding.
+ * @cssproperty --forge-date-time-picker-summary-day-font-size - Summary panel main day-text font size.
+ *
+ * @csspart summary - The optional left-side summary panel (only present when `summary` is set).
```
samrichardsontylertech: This @csspart is a little orphaned up here, splitting the @cssproperty lines apart

packages/forge/src/lib/date-time-picker/date-time-picker.ts
```diff
@@ -0,0 +1,1362 @@
...
+@customElement(DATE_TIME_PICKER_TAG_NAME)
+export class DateTimePickerComponent extends BaseLitElement implements IDateTimePickerComponent {
+  public static styles = unsafeCSS(styles);
+
+  public static formAssociated = true;
+
+  /** @deprecated Used for compatibility with legacy Forge @customElement decorator. */
+  public static [CUSTOM_ELEMENT_NAME_PROPERTY] = DATE_TIME_PICKER_TAG_NAME;
+
+  @property({ attribute: 'time-mode', reflect: true })
```
samrichardsontylertech: You can put the JSDoc comments right on the properties and take the attribute docs out of the component's comment
```ts
/**
 * Whether time selection applies to a single time, a time range, or a preset time option.
 * @attribute time-mode
 * @default single
 */
```

samrichardsontylertech: We're moving away from reflecting attributes, you may want to remove that now and target internal classes for styling. External code/CSS should use [custom states](https://developer.mozilla.org/en-US/docs/Web/API/CustomStateSet) for detecting stuff like disabled status. We have a `toggleState()` utility function to set that.

packages/forge/src/lib/date-time-picker/date-time-picker.ts
```diff
@@ -0,0 +1,1362 @@
...
+  @property({ type: Boolean, reflect: true }) public persistent = false;
+  @property({ attribute: 'placement', reflect: true }) public placement = 'bottom-start';
+
+  @property({ type: Boolean, reflect: true }) public presets = true;
+
+  @state() private _headerEmpty = true;
+  @state() private _footerStartEmpty = true;
+  @state() private _footerCenterEmpty = true;
+  @state() private _footerEndEmpty = true;
+  @state() private _timeLabelEmpty = true;
```
samrichardsontylertech: Same comment here about the `HideWhenEmptyDirective`

packages/forge/src/lib/date-time-picker/date-time-picker.ts
```diff
@@ -0,0 +1,1362 @@
...
+  @state() private _focusedSlotIndex = -1;
```
samrichardsontylertech: We have a focus group directive now that you can probably replace your custom focus handling with

packages/forge/src/lib/date-time-picker/date-time-picker.ts
```diff
@@ -0,0 +1,1362 @@
...
+    this.#updateFormValueAndValidity();
+    this.#warmTemporal();
+    if (typeof window !== 'undefined' && typeof window.matchMedia === 'function') {
+      this.#phoneMql = window.matchMedia('(max-width: 599px)');
+      this.#phoneMql.addEventListener('change', this.#onPhoneChange);
+      this.#onPhoneChange(this.#phoneMql);
+    }
+  }
+
+  /** Lazily loads the Temporal polyfill when `valueMode` needs it, re-rendering once available so the public `value` can resolve. */
```
samrichardsontylertech, line 536: 👍

packages/forge/src/lib/date-time-picker/date-time-picker.ts
```diff
@@ -0,0 +1,1362 @@
...
+    void ensureTemporal().then(() => this.requestUpdate());
+  }
+
+  public override willUpdate(changed: PropertyValues<this>): void {
+    if (changed.has('open') && changed.get('open') !== undefined) {
+      const eventName = this.open ? DATE_TIME_PICKER_CONSTANTS.events.OPEN : DATE_TIME_PICKER_CONSTANTS.events.CLOSE;
+      this.dispatchEvent(new CustomEvent(eventName, { bubbles: true, composed: true }));
+      if (this.open && this.#deferred) {
+        this.#syncFromValue(this.#value);
+        this.#draftValue = this.#value;
```
samrichardsontylertech, line 559: This should probably be extracted to its own method to keep `willUpdate` concise

packages/forge/src/lib/date-time-picker/date-time-picker.ts
```diff
@@ -0,0 +1,1362 @@
...
+      }
+    }
+    if (changed.has('timeMode') && changed.get('timeMode') !== undefined) {
+      const previousMode = changed.get('timeMode') as TimeMode | undefined;
+      if (previousMode && previousMode !== this.timeMode) {
```
samrichardsontylertech: Curious — *can* this value ever match the previous when it's included in the changed values? I actually don't know the answer. Just wondering whether this check is necessary.

packages/forge/src/lib/date-time-picker/date-time-picker.ts
```diff
@@ -0,0 +1,1362 @@
...
+        this.#value = null;
+        this.#activeFromDate = null;
+        this.#activeToDate = null;
+        this.#activeTime = null;
+        this.#activeFrom = null;
+        this.#activeTo = null;
+        queueMicrotask(() => this.#emitChange('mode-change'));
```
samrichardsontylertech: This should probably also be extracted to a new method? Also is `queueMicrotask` the best option here or should this tie into Lit's update cycle?

packages/forge/src/lib/date-time-picker/date-time-picker.ts
```diff
@@ -0,0 +1,1362 @@
...
+    }
+    if (changed.has('valueMode')) {
+      this.#warmTemporal();
+    }
+  }
+
+  public override updated(_changed: PropertyValues<this>): void {
+    this.#updateFormValueAndValidity();
+    // Mirror sheet presentation onto the host so :host can fill the sheet width.
+    this.toggleAttribute('data-sheet', this._isPhone && !!(this.#anchorElement ?? this.anchor));
```
samrichardsontylertech: Custom state would suit this better IMO

packages/forge/src/lib/date-time-picker/date-time-picker.ts
```diff
@@ -0,0 +1,1362 @@
...
+      </aside>
+    `;
+  }
+
+  #renderHeader(): TemplateResult {
+    return html`
+      <div part="header" class="header" data-empty=${String(this._headerEmpty)}>
+        <slot name="header" @slotchange=${(e: Event) => this.#onSlotChange(e, '_headerEmpty')}></slot>
+      </div>
+    `;
```
samrichardsontylertech: Same comment as in the SCSS file — in this and a few of the other render methods you might as well just remove the wrapping `<div>` and use `HideWhenEmptyDirective`

packages/forge/src/lib/date-time-picker/date-time-picker.ts
```diff
@@ -0,0 +1,1362 @@
...
+            `
+          : calendarAndTime}
+      </div>
+    `;
+  }
+
+  #renderPresets(): TemplateResult {
+    return html`
+      <div part="presets" class="presets" role="group" aria-label="Quick date ranges">
+        ${PRESET_DEFS.map(p => html`<button type="button" part="preset" data-preset-id=${p.id} @click=${this.#onPresetClick}>${p.label}</button>`)}
```
samrichardsontylertech: How would you feel about attaching the click listener to the parent element and using event delegation instead of creating one for each preset?

samrichardsontylertech: Also I'd go with Forge buttons here instead of styling your own native buttons

packages/forge/src/lib/date-time-picker/date-time-picker.ts
```diff
@@ -0,0 +1,1362 @@
...
+  }
+
+  #renderSlotList(): TemplateResult {
+    const list = this.#computedSlots();
+    const disabledMap = this.#computedDisabledSlots();
+    const tabIndex = this.#computeActiveTabSlotIndex(list, disabledMap);
+    const labelFmt = this.#getSlotLabelFmt();
+    return html`
+      <div part="slot-list" class="slot-list" role="listbox" aria-label="Available times" aria-orientation="vertical" @keydown=${this.#onSlotListKeydown}>
+        <div class="slot-list-inner" role="presentation">
```
samrichardsontylertech: No need to set `role="presentation"` here, divs are already generic

packages/forge/src/lib/date-time-picker/date-time-picker.ts
```diff
@@ -0,0 +1,1362 @@
...
+  }
+
+  #renderSlot(slot: ITimeSlot, index: number, slotIsDisabled: boolean, activeTabIndex: number, labelFmt: Intl.DateTimeFormat): TemplateResult {
+    const selected = this.#activeTime === slot.value;
+    // Per the ARIA listbox pattern, an unavailable option stays perceivable and focusable
+    // (aria-disabled), so it is NOT natively disabled — only the whole-component disabled/readonly
+    // states remove the buttons from the tab order entirely.
+    const tabIndex = index === activeTabIndex ? 0 : -1;
+    return html`
+      <button
```
samrichardsontylertech: Same comment — Forge buttons would probably work better here

samrichardsontylertech (changes requested):
Sorry for the massive amount of comments! I wasn't able to look through everything but hopefully this is enough to think about for now

packages/forge/src/lib/date-time-field/date-time-field.ts
dannydel: I'm totally fine with allowing the input to be created the user instead. It can always just be a recipe if people are lazy and want copy paste. Let me know your thoughts @samrichardsontylertech

packages/forge/src/lib/date-time-picker/date-time-picker.ts
dannydel: Much better idea, I'll update to do that instead.

PR #1184, opened 2026-07-02
fix: migrate usages of deprecated sass functions

MikeMatusz:
## Summary
Address usages of sass deprecations that I keep seeing logged in builds of applications which use `@tylertech/forge`.  Mostly switching `if()`to `@if`and replacing a few color functions with their new equivalent.

## Checklist
- [ ] Tests added/updated

## Breaking Changes
None

### DIFF
--- a/.changeset/witty-pianos-think.md
+++ b/.changeset/witty-pianos-think.md
@@ -1,0 +1,5 @@
+---
+'@tylertech/forge': patch
+---
+
+fix: migrate usages of deprecated sass functions
--- a/packages/forge/src/lib/calendar/_mixins.scss
+++ b/packages/forge/src/lib/calendar/_mixins.scss
@@ -411,9 +411,13 @@
 }
 
 @mixin date-inner-gap($force: false) {
+  $important: null;
+  @if $force {
+    $important: '!important';
+  }
   position: absolute;
-  width: calc(100% - #{variables.$date-gap}) #{if($force, '!important', '')};
-  height: calc(100% - #{variables.$date-gap}) #{if($force, '!important', '')};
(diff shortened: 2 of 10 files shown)


samrichardsontylertech:
Any concern this could reintroduce the issues that [this PR](https://github.com/tyler-technologies-oss/forge/pull/1124) was made to fix?

MikeMatusz:
> Any concern this could reintroduce the issues that [this PR](https://github.com/tyler-technologies-oss/forge/pull/1124) was made to fix?

It definitely would.  Does Forge align with any kind of support window for Dart Sass versions?  Being on a newer version of sass, an application can definitely get a ton on warnings in the output when using Forge styles.  They're just warnings, so this isn't urgent, but ignoring that many warnings can definitely mask new or more important warnings from our own code which we do want to resolve.

From SASS docs:

> [Dart Sass](https://sass-lang.com/dart-sass) will emit deprecation warnings for at least three months before releasing a breaking change, and will release the breaking change with a new major version number unless that change is necessary for CSS compatibility. CSS compatibility changes are often both non-disruptive and time-sensitive, so they may be released with new minor version numbers instead.

Given that some of the changes are required for compatibility with changes to native CSS, it does seem fair that it might be necessary to expect applications using Forge styles to be on a sufficiently recent version of sass.  1.95.0 was released Dec 8, 2025.  I'm not sure what the most appropriate lifetime would be, but it does seem like with our increased vigilance regarding security, most applications should be trying to stay fairly current with dependencies in general.  Might be worth Forge having a documented stance regarding SASS compatibility.

MikeMatusz:
Related:
- #1196

samrichardsontylertech (approved):

PR #1185, opened 2026-07-06
feat(skeleton): migrated to lit

rapioanfernee:
## Summary
- Converted skeleton component into lit code

## Checklist
- [x] Tests added/updated

## Breaking Changes
None

### DIFF
--- a/packages/forge/src/lib/skeleton/skeleton.ts
+++ b/packages/forge/src/lib/skeleton/skeleton.ts
@@ -1,30 +1,26 @@
-import { customElement, attachShadowTemplate } from '@tylertech/forge-core';
-import { BaseComponent, IBaseComponent } from '../core/base/base-component.js';
+import { CUSTOM_ELEMENT_NAME_PROPERTY } from '@tylertech/forge-core';
+import { customElement, property } from 'lit/decorators.js';
+import { html, TemplateResult, unsafeCSS } from 'lit';
+import { BaseLitElement } from '../core/base/base-lit-element.js';
 import { SKELETON_CONSTANTS } from './skeleton-constants.js';
 
-import template from './skeleton.html';
 import styles from './skeleton.scss';
 
-export interface ISkeletonComponent extends IBaseComponent {}
-
-declare global {
-  interface HTMLElementTagNameMap {
-    'forge-skeleton': ISkeletonComponent;
-  }
+/** @deprecated - This will be removed in the future. Please switch to using SkeletonComponent. */
+export interface ISkeletonComponent extends BaseLitElement {
+  formField: boolean;
+  button: boolean;
+  chip: boolean;
+  listItem: boolean;
+  text: boolean;
+  avatar: boolean;
+  stretch: boolean;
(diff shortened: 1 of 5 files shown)


packages/forge/src/lib/skeleton/skeleton.ts
```diff
@@ -1,31 +1,27 @@
...
 
-export interface ISkeletonComponent extends IBaseComponent {}
-
-declare global {
-  interface HTMLElementTagNameMap {
-    'forge-skeleton': ISkeletonComponent;
-  }
+/** @deprecated - This will be removed in the future. Please switch to using SkeletonComponent. */
+export interface ISkeletonComponent extends BaseLitElement {
+  formField: boolean;
```
rapioanfernee, line 11: Open for suggestions here!

I asked Claude for options for documenting the attributes that are only used by CSS styling (but need to be removed as per migration guideline) and this is one. 

The other one is just adding `@description` in the jsDocs and mentioning the attributes there but I think it does not have the same visibility with `@attribute` as compared to before and with the first option above.

List of attributes:

- form-field
- button
- chip
- list-item
- text
- avatar
- stretch

samrichardsontylertech: That guideline is specifically about removing attribute reflection (meaning the attribute being set every time the property changes). The attributes themselves are fine and can be documented the same way they were before.

Generally you'll want to set a class on the root element of the component and target that for styling instead of the host with the attribute. If the style applies to the host element itself you can set a custom CSS state and target that.

Though I think for this component it's fine to keep the attributes reflected and use them for styling. There's no complex behavior involved and the attribute names add meaning to how the skeleton reads to a dev, kind of like the native `<input type="checkbox" switch />` element.

packages/forge/src/lib/skeleton/skeleton.ts
```diff
@@ -56,12 +52,62 @@ declare global {
...
-    attachShadowTemplate(this, template, styles);
+@customElement(SKELETON_CONSTANTS.elementName)
+export class SkeletonComponent extends BaseLitElement implements ISkeletonComponent {
+  public static styles = unsafeCSS(styles);
+
+  /** @deprecated Used for compatibility with legacy Forge @customElement decorator. */
+  public static [CUSTOM_ELEMENT_NAME_PROPERTY] = SKELETON_CONSTANTS.elementName;
+
+  /**
+   * Applies form field styles to the skeleton.
```
samrichardsontylertech, line 63: You can document the attributes and default values here:
```html
/**
 * Applies form field styles to the skeleton.
 * @default false
 * @attribute form-field
 */
```

rapioanfernee: Applied it on my latest commit: https://github.com/tyler-technologies-oss/forge/pull/1185/commits/d6cfed50844e8530001e0233a9b4dea5131adf69 

Thanks for the review!

samrichardsontylertech (approved):

PR #1205, opened 2026-08-11
feat(listbox): create listbox component

samrichardsontylertech:
## Summary
Adds a `<forge-listbox>` component with drag and drop functionality and declarative `<forge-option>` and `<forge-option-group>` components.

## Checklist
- [x] Tests added/updated

## Breaking Changes
None


Ross-Blakeney:
@Codex review

packages/forge/src/stories/components/listbox/Listbox.mdx
```diff
@@ -0,0 +1,95 @@
...
+## Dragging Between Listboxes
+
+Options can be dragged from one listbox to another. Set `drag-out` on the source listbox and
+`drop-from` on the target listbox with a space-separated list of the source listbox `id`s that
+are allowed to drop into it.
+
+When an option is dropped, the target listbox dispatches a `forge-listbox-drop` event and the
+source listbox dispatches a `forge-listbox-drag-out` event.
+
+<Canvas of={ListboxStories.DragAndDrop} />
```
MikeMatusz, line 59: Can you elaborate on why these are specifically listbox features?

These two examples feel like they'd be more appropriate with a plain list rather than a listbox, mainly because I don't see why a user would want to both reorder a list and select one item or for the latter, dragging to another container would typically be in lieu of selection, so selection feels redundant.

Unless, maybe, the selection is exclusively meant to back keyboard accessibility for the reorder/list-builder operations themselves?  Like being able to multi-select items and move them together with a keyboard shortcut or buttons.  Though those operations don't seem to have keyboard support yet.

samrichardsontylertech: I followed [this pattern from WCAG](https://www.w3.org/WAI/ARIA/apg/patterns/listbox/examples/listbox-rearrangeable/). It uses the selections for moving options but there's nothing stopping other actions from working on the selected options.

I haven't built keyboard shortcuts for this stuff into the component because each listbox doesn't really have a concept of what else is on the page and where it is spatially (whether it's to the left or right of another listbox). These aren't really actions that are native to the listbox spec either, they're more dependent on external buttons for an accessible alternative. In the WCAG example the shortcuts activate the buttons instead of doing anything directly to the listboxes.

Once the basic components are merged in I'm going to add an extended component that combines two listboxes with the buttons and keyboard shortcuts for this specific use case.

And yeah, I'm in total agreement that drag and drop isn't really a listbox-specific thing. I built the drag and drop controllers to be generalized utilities that we can add to any component down the line, including the list. This was just the first one to use them.

packages/forge/src/lib/listbox/listbox.ts
```diff
@@ -0,0 +1,864 @@
...
+import { DragController } from '../core/controllers/drag-controller.js';
+import { DropController, DropEventArgs } from '../core/controllers/drop-controller.js';
+import { setDefaultAria } from '../core/utils/a11y-utils.js';
+import { composedPathFrom } from '../core/utils/event-utils.js';
+import { FocusGroupController } from '../core/utils/focus-group.js';
+import { FormRestoreReason, FormRestoreState } from '../core/utils/form-utils.js';
+import { KeyActionController } from '../core/utils/key-action.js';
+import { toggleState } from '../core/utils/utils.js';
+import type { OptionGroupComponent } from '../option/option-group/index.js';
+import type { OptionComponent, OptionUpdateReason } from '../option/option/index.js';
```
DRiFTy17, line 18: Do we want to include side effect imports here for option and option-group?
```
import '../option/option/option.js';
import '../option/option-group/option-group.js';
```

packages/forge/src/lib/core/controllers/drag-controller.ts
```diff
@@ -0,0 +1,262 @@
...
+    return this.#config.getDragItem?.({ draggableElement, event }) ?? draggableElement;
+  }
+
+  #getDefaultDragImage(event: DragEvent, dragItem: HTMLElement): DragImageConfig {
+    return {
+      image: dragItem,
+      x: event.offsetX,
+      y: event.offsetY
+    };
+  }
```
DRiFTy17, line 238: The drag image looks to be offset, I was able to fix it locally by switching to this:
```typescript
  #getDefaultDragImage(event: DragEvent, dragItem: HTMLElement): DragImageConfig {
    const rect = dragItem.getBoundingClientRect();
    return {
      image: dragItem,
      x: event.clientX - rect.left,
      y: event.clientY - rect.top
    };
  }
```

packages/forge/src/lib/option/option/option.ts
```diff
@@ -0,0 +1,278 @@
...
+  @state()
+  private _focusIndicatorActive = false;
+
+  @query('forge-state-layer', true)
+  private _stateLayer!: StateLayerComponent;
+
+  #internals: ElementInternals;
+  #configOnly = true;
+
+  constructor() {
```
DRiFTy17, line 154: I added the following listeners here to remove the "pressed" state on the drag image while dragging and it seemed to work:
```typescript
this.addEventListener('dragstart', () => toggleState(this.#internals, 'dragging', true));
this.addEventListener('dragend', () => toggleState(this.#internals, 'dragging', false));
```

Then in `option.scss` I added the following:
```scss
:host(:state(dragging)) forge-state-layer {
  display: none;
}
```

packages/forge/src/lib/core/controllers/drag-controller.ts
samrichardsontylertech: That didn't fix anything for me — Chrome is still broken and Safari still works

packages/forge/src/stories/components/listbox/Listbox.mdx
MikeMatusz: I agree that the list-builder operations are outside the scope of the listbox itself, bur re-ordering within a single listbox, especially if supported intrinsically via drag and drop, potentially should also have intrinsic keyboard support.  Maybe this example would make more sense in the context of a list builder or other pattern, or if multi-selecting allowed you to drag and drop multiple items together.  Just felt odd when I was using it because I couldn't imagine a use case for the particular combination of features.

samrichardsontylertech: @MikeMatusz I added in a QOL compromise with built in `shiftSelectionOptionUp()` and `shiftSelectedOptionDown()` methods and support for invoker commands —
```html
<forge-listbox id="listbox" reorderable>...</forge-listbox>
<forge-button command="--shift-up" commandfor="listbox">Move options up</forge-button>
<forge-button command="--shift-down" commandfor="listbox">Move options down</forge-button>
```
With an enhancement to the keyboard shortcut component to make activating the buttons from the listbox easier

PR #1228, opened 2026-09-08
feat(app-bar-menu-button): add ARIA attributes to show menu button state

angelalim-tyler:
## Summary
Updates will add `aria-expanded`, `aria-haspopup` and `aria-controls` on the menu-button of forge-app-bar component. This enhancement will let screen readers announce the state of the menu (whether expanded or collapsed). Developers using this component does not have to add the ARIA attributes manually as this would create issues flagged by acceesibility tools.

## Checklist
- [x] Tests added/updated

## Breaking Changes
None

### DIFF
--- a/.changeset/hot-rooms-buy.md
+++ b/.changeset/hot-rooms-buy.md
@@ -1,0 +1,5 @@
+---
+'@tylertech/forge': minor
+---
+
+feat(app-bar-menu-button): add `expanded`, `controls`, and `hasPopup` properties for ARIA state on the internal button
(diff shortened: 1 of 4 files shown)


nickonometry:
The PR looks fine to me, but just one general comment. I wonder if we should just convert this component to a Lit component while we're here?

samrichardsontylertech:
Thanks for making this PR! It makes sense to me, though I'm a little hesitant to create new properties for ARIA. Ideally we'd make the component host element the semantic button so it would support ARIA natively. We do this in the other button component and probably should here.

samrichardsontylertech (approved):
After discussing this, we think the changes are good to merge 👍

We decided to deprecate the menu button in favor of just using the icon button component so this is a temporary, but needed, fix.

PR #1233, opened 2026-09-10
feat(menu): add onSelect callback to menu options

shoebox639:
## Summary

Adds an optional `onSelect` callback to `IMenuOption`. When an option is selected, the menu calls it with the same `{ index, value }` data it dispatches on `forge-menu-select`.

Today a menu that lists actions needs one handler that branches over every option's `value`, so the option definition and its behavior live in different places and nothing checks that every value has a branch. With `onSelect`, an action menu is a list of options that each carry their own handler:

```ts
const options: IMenuOption[] = [
  { label: 'Print', value: 'print', onSelect: () => print() },
  { label: 'Export', value: 'export', onSelect: () => exportReport() }
];
```

Behavior:

- Runs after `forge-menu-select` is dispatched. If a listener calls `preventDefault()`, the selection is cancelled and the callback does not run, matching how `selectedValue` already behaves.
- For cascading menus, a nested option's callback is called once, by the child menu that renders it. The parent's re-dispatched event with `parentValue` is unchanged.
- Menus without callbacks behave exactly as before. The `options` setter's shallow copy already preserves function references, so the callback survives it and the child-menu copy.

Testing: four new tests in `menu.test.ts` cover the happy path, ordering relative to the event, cancellation, and child-menu options. `pnpm run --filter @tylertech/forge test src/lib/menu/` passes (72 tests). Package build, eslint, and prettier are clean.

## Checklist
- [x] Tests added/updated

## Breaking Changes
None


shoebox639:
@DRiFTy17 @samrichardsontylertech I wanted to call this out:

> [!WARNING]
> This `onSelect` Runs AFTER forge-menu-select is dispatched. If a listener calls preventDefault(), the selection is cancelled and the callback does not run, matching how selectedValue already behaves. 

Maybe we should do this before, and let the user prevent propagation at this level rather than the other way around?

packages/forge/src/lib/menu/menu-core.ts
```diff
@@ -441,6 +441,7 @@ export class MenuCore extends CascadingListDropdownAwareCore<IMenuOption | IMenu
     const isCancelled = !this._adapter.emitHostEvent(MENU_CONSTANTS.events.SELECT, data, true, true);
     if (!isCancelled) {
       this._selectedValue = this._persistSelection ? value : undefined;
+      onSelect?.(data);
```
DRiFTy17, line 444: I'm fine with letting the event bubble before calling `onSelect`, but the thought crossed my mind, I do wonder if it should run regardless of event propagation?

If we do keep it gated by event cancelation, then I think we need to also handle the case where a child menu option select is event is cancelled when using a cascading menu. I think right now the child menu select event doesn't get canceled if a parent select event does.

shoebox639: I guess it comes down to what a dev would expect `onSelect: (e) => e.stopPropagation()` when combined with `on-forge-menu-select=...` to do? 

Personally I would expect `on-forge-menu-select` not fire for this specific option. But maybe that's just because I haven't really used this much and that pattern hasn't stuck yet.

DRiFTy17: I think I would avoid passing the event as an arg to `onSelect`. This new callback seems like something that should be separate from events altogether IMO. I think it would be confusing to tie the two together because if you're using `onSelect` then you're likely not using the `forge-menu-select` event anyway...

The typical pattern is to use native DOM events for this type of thing. I think the difference here, and likely the main driver for your proposal, is that `<forge-menu>` doesn't have declarative `<forge-menu-option>` elements in the DOM. If it did, you could listen for the select event on an individual menu option element. This is a pattern that Forge is moving to for dropdown-based components, but the menu hasn't been refactored yet.

samrichardsontylertech: I'm actually torn on this. The callback should run when the option is selected, but if you call `preventDefault()` on the event the selection doesn't actually happen. It's unlikely to come up in practice because no one will use the event and the callback together. I'd follow the native behavior for this — how does event cancellation work with `HTMLElement.onClick()`?

Kieran raises an important point though — we're moving to declarative option elements so before too long the preferred pattern will be `<forge-option (click)="doSomething()">` anyway.

samrichardsontylertech (approved):

manojsridharsrinivasan:
@samrichardsontylertech @shoebox639 This PR had a ripple effect and broke our applications. We have fixed it with https://github.com/tyler-technologies/energov-workflow-angular-ui/pull/405 but just wanted to let you know in case if it is worth feel revisiting. Please look at `item-review-corrections-edit.component.ts` changes.

PR #1244, opened 2026-09-14
feat(keyboard-shortcut): add shortcut scoping, sequence shortcuts, and api alignment

Alanoll:
## Summary
Extends `forge-keyboard-shortcut` so shortcuts can be scoped to a region of the page instead of only "one element" or "the whole   document." Today, two independent widgets on the same page (e.g. two editors) can't each bind the same shortcut to their own controls without one triggering the other. 

This enhancement provides a means to define that scope via `forge-keyboard-shortcut-scope` in the nearest parent enclosing the shortcut, so the same key can safely exist in multiple places at once, with whichever region has focus taking priority.

## Changes
 * Unblocks any page layout with repeated or nested interactive regions that each want their own shortcuts (multiple editors, panels, or widgets sharing common shortcuts like Bold/Save/Undo, or modals).
 * Removes a class of bugs where enabling a shortcut in one part of a page unintentionally affects another.
 * Adds accessibility improvements (shortcuts are now announced to assistive tech automatically) and platform-aware key handling (Cmd on Mac, Ctrl elsewhere), reducing one-off work every consuming team previously had to do by hand.
 * Adds support for multi-key shortcut sequences (e.g. "g then i"), broadening the range of shortcut patterns Forge can support.

The changes _should_ be backwards compatible. The legacy `target` property is still present and maps it's results to the two new properties `anchor` and `scope`.

The `anchor` is the tranditional "target" of the shortcut, but I went with the name `anchor` to match other Forge elements (`focus-indicator` and `popover`). It's the element that will have Aria attributes added. Accepts the ID of the element to become the anchor. A matching `anchorElement` allows for explicit assignment of an element.

The `scope` is the element that limits where the shortcut can be trigger from. The original implementation had either the target element itself must have focus or be global. `global` is still available and will set the `scope` to the document itself. Accept the ID of the element to become the scope. A matching `scopeElement` allows for explicit assignment of an element.

`allowRepeat` will only held down keys to fire repeated events (e.g. holding the arrow key to move a canvas in small increments). If false, will only fire once.

## Docs
I updates the DEV pages webapp with example, and in the process writing the Storybook docks. I'll flip this from Draft when that's done.

## Checklist
- [X] Tests added/updated
- [In Progress ] Docs updated (if applicable)
- [X] Changeset added (`pnpm changeset`)

## Breaking Changes
There _shouldn't_ be breaking changes. Everything should be backwards compatible. I thought about changing the `event.detail`, but stuck with non-breaking changes.


Alanoll:
@samrichardsontylertech Could you give it a once over to see if I missed something from a fundamental `Forge` coding standard? I mean look at everything else, but just to make sure I'm not causing something to go backwards from the general Forge plan.

Alanoll:
@samrichardsontylertech I think it's ready for review. I'm not sure what needs to be done for the Angular proxies to get updated though (especially now that they're in the same monorepo).

samrichardsontylertech:
@Alanoll  This is next on my list 👍

I have a `<forge-kbd>` component PR up as well so I'm going to try to keep the API as similar as possible between these two

PR #1247, opened 2026-09-16
feat(process-stepper): add process stepper component

sriramasdanturthi:
## Summary

Adds `<forge-process-stepper>` and `<forge-process-step>` — an interactive stepper that shows where a user is within a multi-step process while allowing actions to be completed directly from each step. Implements [Process Stepper](https://tylertech.atlassian.net/wiki/spaces/TF/pages/1477355585/Process+Stepper) (FORGE-31).

Unlike `<forge-stepper>`, a process step can contain arbitrary interactive content, so users can complete tasks, acknowledge information, launch a workflow, or navigate to a related page without leaving the page the stepper is on. The existing `<forge-stepper>` is untouched.

### One component, two behaviours

Rather than a `default` and an `interactive` variant, interactivity is derived from content, the way `forge-list-item` does it: the step's label is slotted, and a link or button in that slot becomes the step's interactive area. A step whose label is plain text stays a read-only indicator of progress.

```html
<forge-process-stepper numbered>
  <span slot="title">Record progress</span>

  <!-- read-only -->
  <forge-process-step state="completed">
    Application received
    <span slot="meta">Started:</span>
    <span slot="meta">01/15/2026</span>
  </forge-process-step>

  <!-- interactive: navigates -->
  <forge-process-step state="completed"><a href="/review">Internal review</a></forge-process-step>

  <!-- interactive: acts, with content completed in place -->
  <forge-process-step state="current">
    <button>Suspension</button>
    <forge-checkbox slot="additional-content"><label>Issue suspension notice</label></forge-checkbox>
    <forge-button slot="actions" variant="outlined">Advance to next stage</forge-button>
  </forge-process-step>
</forge-process-stepper>
```

A slotted link wins over a slotted button. Only the label slot is considered, so a button in `actions` or `additional-content` does not make the step itself interactive, and `noninteractive` opts out. This also means a step can genuinely navigate — with an `href`, middle-click, and open-in-new-tab — which an internal button could never do.

### Behaviour

- **Orientations** — `vertical` (default) and `horizontal`, covering both requesting products.
- **Progress** — the connecting line doubles as the progress indicator: a rail beside the markers when vertical, a track above them when horizontal, filled through the `completed`/`current`/`in-progress` steps so it reads as a bar that fills as the process advances. `progress` returns the completed fraction.
- **Small screens** — below 600px of available width a horizontal stepper lays its steps out vertically, because four steps in a 360px container leaves roughly 48px per label. Driven by the stepper's own width via `ForgeResizeObserver`, so a stepper in a narrow column adapts too; `orientation` is never mutated and the read-only `compact` property reports the state.
- **States** — all 11 states from the spec, each with a generated marker treatment (filled + check, partially filled, dashed outline, error outline + exclamation).
- **Step numbers** — markers are empty by default; `numbered` on the stepper renders each step's number, matching the two variants in the design.
- **Slots** — `title` on the stepper; `marker`, `meta`, `message`, `actions`, `additional-content` and the label slot on each step. `meta` lays out in two columns so label/value pairs stay aligned down the step.

Geometry was measured against the design export rather than eyeballed: 8px line-to-marker gap, 24px horizontal step spacing, uniform 72px vertical rhythm, circular markers at every state and size, and a continuous progress track.

### Accessibility (WCAG 2.2 AA)

Each step is a single list item: `role="list"` on the stepper, `role="listitem"` on each step, with `aria-posinset`, `aria-setsize`, and `aria-current="step"` on the current step. An interactive step contains the consumer's own link or button as its interactive area, and the description, meta content, messaging, and step content are siblings of it rather than nested inside, so interactive content keeps its own semantics. Markers are decorative and `aria-hidden`.

A `<forge-focus-indicator>` draws the focus ring around the label text only, so keyboard focus never outlines the progress line, the label's full cell, or the step's content. The label slot is wrapped in an element that shrink-wraps its content, matching the `step-label` grouping in the design's layer structure.

Per design review, the ring sits further from the text along the inline axis (8px) than the block axis (4px) so short labels are not crowded, and has a 4px corner radius. All three are configurable via `--forge-process-step-focus-indicator-offset`, `-offset-block`, and `-shape`. Two consequences worth a design call: at the default inline offset the ring meets the marker exactly in the vertical orientation, and it extends 8px past the leading edge of a horizontal stepper. Both follow from an outset ring; switching to the focus indicator's `inward` variant instead is a one-line change if that reads better.

A polite live region announces the step a process moves to — "Step 3 of 5: Internal review" — whenever the `current` step changes, including when a consumer advances the process programmatically. It stays silent on the initial render.

The verified accessibility tree for an interactive step with content:

```
list
  listitem
    button "Suspension"
    checkbox "Issue suspension notice"
    button "Advance to next stage"
```

### Also included

- Design tokens under `core/styles/tokens/process-stepper/` (26 CSS custom properties, 10 CSS parts on the step)
- Storybook docs (`ProcessStepper.mdx`) with 10 stories
- A dev-site page at `/pages/process-stepper/process-stepper.html`, registered in the component list
- Registered in `lib/index.ts` (`defineProcessStepperComponent`, `defineProcessStepComponent`, and `defineComponents()`)

## Checklist
- [x] Tests added/updated — 57 tests in `process-stepper.test.ts`, all passing; full suite 3872 passed / 0 failed
- [x] Docs updated (if applicable) — `ProcessStepper.mdx` plus generated API from the manifest
- [x] Changeset added (`pnpm changeset`) — `.changeset/process-stepper.md` (minor)

## Breaking Changes

None. Both elements are new, and `<forge-stepper>` / `<forge-step>` are unchanged. One commit is marked `!` because it removed `clickable` and the `label` attribute from this branch's own earlier commit; nothing has been released, so there is no consumer impact.

## Notes for reviewers

- **Naming** — `forge-process-stepper` rather than taking the `forge-stepper` name, so the legacy stepper needs no deprecation. The spec's open question about renaming the old one to `oldStepper` is deliberately left alone.
- **Vertical rhythm** — the uniform row height is `--forge-process-step-row-min-block-size: 72px`, derived from the design export at an inferred ~2.1x scale. If the source marker is 24px rather than 20px the intended value is closer to 82px; worth confirming against Figma.
- **Label markup** — the label sits in its own grid cell rather than inside `part="content"`, so a marker larger than the label (for example a slotted avatar) can share a row with it and stay level with the start of the progress line.
- **Bold labels** — the "Stage progress" example in the design uses bold step labels while the other vertical example does not. Labels are regular weight here; if bold is intended for that pattern it needs a token or property.
- **Docs page CSS** — `ProcessStepper.mdx` carries a small scoped style block and a `toc` parameter. Both work around Storybook's generated API tables: long type names such as `CustomEvent<IProcessStepperChangeEventData>` render in a `nowrap` `code` element and overflow a narrow docs pane, and an `h4` per table made the contents list long enough to scroll and clip. `forge-table` has the same overflow and is untouched.
- **Not included** — the legacy stepper's `fixed`/`clustered` layout modes (horizontal steps distribute evenly), and arrow-key roving focus. Interactive steps use the consumer's own link or button, which is WCAG-conformant, though arrow-key navigation is a conventional addition for a horizontal stepper and would be a reasonable follow-up.
- **Still open from the spec** — parity with the prior stepper, the planned stepper rebuild, the relationship to the "gauge" component, and whether this should be built on the timeline component.


packages/forge/src/lib/process-stepper/process-stepper/process-stepper.ts
```diff
@@ -0,0 +1,228 @@
...
+
+  /* @internal */
+  public render(): TemplateResult {
+    return html`
+      <div part="root" class=${classMap({ 'forge-process-stepper': true, [this.#effectiveOrientation]: true, compact: this.compact })}>
+        <div part="title" class="title" ${hideWhenEmpty()}><slot name="title"></slot></div>
+        <div part="steps" class="steps">
+          <slot @slotchange=${this.#handleSlotChange}></slot>
+        </div>
+        <div part="announcer" class="announcer" role="status" aria-live="polite">${this._announcement}</div>
```
samrichardsontylertech: You can remove this and use the shared live announcer from Forge Core

packages/forge/src/lib/process-stepper/process-stepper/process-stepper.ts
```diff
@@ -0,0 +1,228 @@
...
+    }
+    const completed = steps.filter(step => step.state === 'completed').length;
+    return completed / steps.length;
+  }
+
+  /* @internal */
+  public render(): TemplateResult {
+    return html`
+      <div part="root" class=${classMap({ 'forge-process-stepper': true, [this.#effectiveOrientation]: true, compact: this.compact })}>
+        <div part="title" class="title" ${hideWhenEmpty()}><slot name="title"></slot></div>
```
samrichardsontylertech: We typically leave content like titles out of core components in favor of allowing them to be composed externally (e.g. `<h3 class="forge-typography-subheading1>Title</h3><forge-process-stepper>...</forge-process-stepper>`)

packages/forge/src/lib/process-stepper/process-stepper/process-stepper.ts
```diff
@@ -0,0 +1,228 @@
...
+
+  /** @deprecated Used for compatibility with legacy Forge @customElement decorator. */
+  public static [CUSTOM_ELEMENT_DEPENDENCIES_PROPERTY] = [ProcessStepComponent];
+
+  /**
+   * The orientation of the process.
+   * @default 'vertical'
+   * @attribute
+   */
+  @property({ reflect: true })
```
samrichardsontylertech: You can remove attribute reflection — it's going away across the library

packages/forge/src/lib/process-stepper/process-stepper/process-stepper.ts
```diff
@@ -0,0 +1,228 @@
...
+  public orientation: ProcessStepperOrientation = 'vertical';
+
+  /**
+   * Whether each step displays its number within the marker. Steps with an icon marker, such as
+   * completed and error steps, always show the icon instead.
+   * @default false
+   * @attribute
+   */
+  @property({ type: Boolean, reflect: true })
```
samrichardsontylertech: You can remove this attribute reflection too

packages/forge/src/lib/process-stepper/process-stepper/process-stepper.ts
```diff
@@ -0,0 +1,228 @@
...
+    }
+
+    this.#announcedIndex = index;
+
+    if (index === -1) {
+      return;
+    }
+
+    const step = steps[index];
+    const position = `Step ${index + 1} of ${steps.length}`;
```
samrichardsontylertech: The announced position could be incorrect if `aria-posinset` or `aria-setsize` are in use, in which case they should take precedence over the DOM position and elements

packages/forge/src/lib/process-stepper/process-stepper/process-stepper.ts
```diff
@@ -0,0 +1,228 @@
...
+  #onStepSelect: EventListener = (evt: Event) => this.#handleStepSelect(evt);
+
+  #handleSlotChange(): void {
+    this.#syncSteps();
+  }
+
+  #syncSteps(): void {
+    const steps = this._steps;
+    steps.forEach((step, i) => {
+      step.index = i + 1;
```
samrichardsontylertech: Same here with `aria-posinset`. Should that be used for the step number?

samrichardsontylertech: Or, alternatively, should `index` be part of the step's public API and determine the value here and/or of `aria-posinset`?

packages/forge/src/lib/process-stepper/process-stepper/process-stepper.ts
```diff
@@ -0,0 +1,228 @@
...
+      step.count = steps.length;
+      step.last = i === steps.length - 1;
```
samrichardsontylertech: Same here with `aria-setsize`?

packages/forge/src/lib/process-stepper/process-stepper/process-stepper-constants.ts
```diff
@@ -0,0 +1,36 @@
...
+import type { ProcessStepComponent } from '../process-step/process-step.js';
+
+const elementName: keyof HTMLElementTagNameMap = `${COMPONENT_NAME_PREFIX}process-stepper`;
+
+const attributes = {
+  ORIENTATION: 'orientation'
+};
+
+const events = {
+  CHANGE: `${elementName}-change`
```
samrichardsontylertech: This is unused. That said, I'd prefer if the stepper emitted a `change` event instead of a `forge-process-stepper-change` event. We're trying to move away from custom events in favor of native events with lighter payloads. Getting the current step could look like
```typescript
stepper.addEventListener('change', (event: Event) => {
  console.log(event.target.currentStep);
}
```

packages/forge/src/lib/process-stepper/process-step/process-step.ts
```diff
@@ -0,0 +1,322 @@
...
+    const interactive = this.interactive;
+
+    return html`
+      <div part="root" class=${classMap({ 'forge-process-step': true, [this.orientation]: true, [this.state]: true, interactive })}>
+        <div part="line" class=${classMap({ line: true, active: this.lineActive, last: this.last })} aria-hidden="true"></div>
+        <div class="sidebar">${this.#marker}</div>
+        <div part="label" class="label-row">
+          <span class="label">
+            <slot @slotchange=${this.#handleLabelSlotChange}></slot>
+            ${when(interactive, () => html`<forge-focus-indicator part="focus-indicator"></forge-focus-indicator>`)}
```
samrichardsontylertech: Include a `<forge-state-layer>` too, for hover and pressed states

packages/forge/src/lib/process-stepper/process-step/process-step.ts
```diff
@@ -0,0 +1,322 @@
...
+          </span>
+        </div>
+        <div part="content" class="content">
+          ${when(this.description, () => html`<span part="description" class="description">${this.description}</span>`)}
+          <div part="meta" class="meta" ${hideWhenEmpty()}><slot name="meta"></slot></div>
+          <div part="message" class="message" ${hideWhenEmpty()}><slot name="message"></slot></div>
```
samrichardsontylertech: There isn't a `message` class in the CSS so this doesn't render any differently than the additional content. I'm torn on whether these separate slots are actually needed anyway? Maybe one multipurpose content slot is enough to cover all the use cases

packages/forge/src/lib/process-stepper/process-step/process-step.ts
```diff
@@ -0,0 +1,322 @@
...
+    this.requestUpdate();
+  }
+
+  #handleClick(evt: Event): void {
+    const interactiveElement = this.#interactiveElement;
+    if (!interactiveElement || !evt.composedPath().includes(interactiveElement)) {
+      return;
+    }
+
+    this.dispatchEvent(new CustomEvent('forge-process-step-select', { bubbles: true, composed: true }));
```
samrichardsontylertech: No need for this to be a custom event — it doesn't include a detail object. `new Event` will do the job.

packages/forge/src/lib/process-stepper/process-step/process-step.ts
```diff
@@ -0,0 +1,322 @@
...
+  }
+}
+
+declare global {
+  interface HTMLElementTagNameMap {
+    'forge-process-step': ProcessStepComponent;
+  }
+
+  interface HTMLElementEventMap {
+    'forge-process-step-select': CustomEvent<void>;
```
samrichardsontylertech: Same here — `'forge-process-step-select': Event;`

packages/forge/src/lib/process-stepper/process-step/process-step.ts
```diff
@@ -0,0 +1,322 @@
...
+  static {
+    IconRegistry.define([tylIconCheck, tylIconExclamation]);
+  }
+
+  /**
+   * The state of the step within the process.
+   * @default 'not-started'
+   * @attribute
+   */
+  @property({ reflect: true })
```
samrichardsontylertech: You can remove attribute reflection from all these properties

packages/forge/src/lib/process-stepper/process-step/process-step.ts
```diff
@@ -0,0 +1,322 @@
...
+  @property({ type: Boolean, reflect: true })
+  public noninteractive = false;
+
+  /**
+   * The one-based position of the step within the process. Set by the parent
+   * `<forge-process-stepper>`.
+   * @default 0
+   */
+  @property({ type: Number, attribute: false })
+  public index = 0;
```
samrichardsontylertech: If this is only intended to be set from the stepper it should probably be removed from the public API. You can create a symbol to keep it internal.

packages/forge/src/lib/process-stepper/process-step/process-step.ts
```diff
@@ -0,0 +1,322 @@
...
+
+  /**
+   * Whether the step number is displayed within the marker when the step has no icon treatment.
+   * Set by the parent `<forge-process-stepper>`.
+   * @default false
+   */
+  @property({ type: Boolean, attribute: false })
+  public numbered = false;
```
samrichardsontylertech: Same comment here about exposing this to the public API. In this case consider using Lit context to subscribe to the value from the stepper.

packages/forge/src/lib/process-stepper/process-step/process-step.ts
```diff
@@ -0,0 +1,322 @@
...
+
+  /**
+   * The total number of steps in the process. Set by the parent `<forge-process-stepper>`.
+   * @default 0
+   */
+  @property({ type: Number, attribute: false })
+  public count = 0;
```
samrichardsontylertech: This is another good property to consider replacing with context

packages/forge/src/lib/process-stepper/process-step/process-step.ts
```diff
@@ -0,0 +1,322 @@
...
+   */
+  @property({ type: Boolean, attribute: false })
+  public last = false;
+
+  /**
+   * The orientation inherited from the parent `<forge-process-stepper>`.
+   * @default 'vertical'
+   */
+  @property({ attribute: false })
+  public orientation: ProcessStepperOrientation = 'vertical';
```
samrichardsontylertech: Same comment here regarding the public API and context

packages/forge/src/lib/core/styles/tokens/process-stepper/process-step/_tokens.scss
```diff
@@ -0,0 +1,45 @@
...
+  content-padding-inline-end: utils.module-val(process-step, content-padding-inline-end, spacing.variable(large)),
+  content-padding-block-end: utils.module-val(process-step, content-padding-block-end, spacing.variable(medium)),
+  row-min-block-size: utils.module-val(process-step, row-min-block-size, 72px),
+  marker-background: utils.module-val(process-step, marker-background, theme.variable(surface)),
+  disabled-opacity: utils.module-val(process-step, disabled-opacity, 0.38),
+  label-content-gap: utils.module-val(process-step, label-content-gap, spacing.variable(xsmall)),
+  marker-label-gap: utils.module-val(process-step, marker-label-gap, spacing.variable(xsmall)),
+  focus-indicator-offset: utils.module-val(process-step, focus-indicator-offset, spacing.variable(xxsmall)),
+  focus-indicator-offset-block: utils.module-val(process-step, focus-indicator-offset-block, spacing.variable(xxxsmall)),
+  focus-indicator-shape: utils.module-val(process-step, focus-indicator-shape, shape.variable(medium))
```
samrichardsontylertech: No need to create these tokens, the focus indicator's native tokens still apply and are accessible to consumers

packages/forge/src/lib/core/styles/tokens/process-stepper/process-step/_tokens.scss
```diff
@@ -0,0 +1,45 @@
...
+$tokens: (
+  marker-size: utils.module-val(process-step, marker-size, 20px),
+  marker-row-block-size: utils.module-val(process-step, marker-row-block-size, typography.variable(body2, line-height)),
+  marker-shape: utils.module-val(process-step, marker-shape, shape.variable(full)),
+  marker-color: utils.module-val(process-step, marker-color, theme.variable(text-high)),
+  marker-icon-color: utils.module-val(process-step, marker-icon-color, theme.variable(surface)),
+  marker-inactive-color: utils.module-val(process-step, marker-inactive-color, theme.variable(text-low)),
+  marker-border-width: utils.module-val(process-step, marker-border-width, spacing.variable(xxxsmall)),
+  marker-error-color: utils.module-val(process-step, marker-error-color, theme.variable(error)),
+  line-width: utils.module-val(process-step, line-width, 3px),
```
samrichardsontylertech: `border.variable(medium)` or `spacing.variable(xxxsmall)` may be a good fit here

packages/forge/src/lib/core/styles/tokens/process-stepper/process-step/_tokens.scss
```diff
@@ -0,0 +1,45 @@
...
+  line-active-width: utils.module-val(process-step, line-active-width, 3px),
```
samrichardsontylertech: By default this is probably intended to mirror the inactive width — `utils.module-ref(process-step, line-active-width, line-width)` will do that even if `line-width` changes

packages/forge/src/lib/core/styles/tokens/process-stepper/process-step/_tokens.scss
```diff
@@ -0,0 +1,45 @@
...
+  line-color: utils.module-val(process-step, line-color, theme.variable(outline-low)),
+  line-active-color: utils.module-val(process-step, line-active-color, theme.variable(text-high)),
+  line-gap: utils.module-val(process-step, line-gap, spacing.variable(xsmall)),
+  sidebar-gap: utils.module-val(process-step, sidebar-gap, spacing.variable(small)),
+  content-gap: utils.module-val(process-step, content-gap, spacing.variable(xxsmall)),
+  meta-gap: utils.module-val(process-step, meta-gap, spacing.variable(xxsmall)),
+  meta-column-gap: utils.module-val(process-step, meta-column-gap, spacing.variable(small)),
+  control-inset: utils.module-val(process-step, control-inset, 12px),
```
samrichardsontylertech: This corresponds to `spacing.variable(small)`

samrichardsontylertech (changes requested):
This is a good first pass! I left some thoughts and shared it with the designers for some extra feedback

packages/forge/src/lib/process-stepper/process-stepper/process-stepper.ts
sriramasdanturthi: Done. Removed the component's own live region; the step now announces itself through `LiveAnnouncer.instance` when it becomes current.

packages/forge/src/lib/process-stepper/process-stepper/process-stepper.ts
sriramasdanturthi: Removed the `title` slot, part and token; the docs and stories now use an external heading linked with `aria-labelledby`.

packages/forge/src/lib/process-stepper/process-stepper/process-stepper.ts
sriramasdanturthi: Removed.

packages/forge/src/lib/process-stepper/process-stepper/process-stepper.ts
sriramasdanturthi: Removed.

packages/forge/src/lib/process-stepper/process-stepper/process-stepper.ts
sriramasdanturthi: Good catch. A consumer's `aria-posinset`/`aria-setsize` now take precedence in the announcement, and they're no longer overwritten because positional ARIA now goes only through ElementInternals.

packages/forge/src/lib/process-stepper/process-stepper/process-stepper.ts
sriramasdanturthi: Yes, the marker number now uses `aria-posinset` when it's set and falls back to the DOM position.

sriramasdanturthi: On making `index` public: I kept `index` internal (a symbol) and made `aria-posinset` the public override, so there's only one source of truth.

packages/forge/src/lib/process-stepper/process-stepper/process-stepper.ts
sriramasdanturthi: Same here: `aria-setsize` overrides the DOM count for both the announcement and the accessibility tree.

packages/forge/src/lib/process-stepper/process-stepper/process-stepper-constants.ts
sriramasdanturthi: Switched to a native `change` Event and removed the custom event and its detail interface. I named the getter `selectedStep` rather than `currentStep`, since `current` is already a consumer-owned step state and activating a step doesn't change it; happy to rename if you prefer.

packages/forge/src/lib/process-stepper/process-step/process-step.ts
sriramasdanturthi: Added a `<forge-state-layer>` (exposed as the `state-layer` part) that targets the slotted link or button and fills the area inside the focus ring.

packages/forge/src/lib/process-stepper/process-step/process-step.ts
sriramasdanturthi: Removed `message` because it rendered the same as `additional-content`; I kept `meta` (two-column grid) and `actions` (wrapping row) because each has its own layout, but can fold them in too if you'd prefer a single slot.

packages/forge/src/lib/process-stepper/process-step/process-step.ts
sriramasdanturthi: Done, it's now a plain `new Event(...)`.

packages/forge/src/lib/process-stepper/process-step/process-step.ts
sriramasdanturthi: Done.

packages/forge/src/lib/process-stepper/process-step/process-step.ts
sriramasdanturthi: Removed from all properties; the disabled style now uses `:state(disabled)`.

packages/forge/src/lib/process-stepper/process-step/process-step.ts
sriramasdanturthi: Moved to an internal symbol that the stepper sets.

packages/forge/src/lib/process-stepper/process-step/process-step.ts
sriramasdanturthi: Now comes from the stepper through Lit context.

packages/forge/src/lib/process-stepper/process-step/process-step.ts
sriramasdanturthi: Now comes from the stepper through Lit context.

packages/forge/src/lib/process-stepper/process-step/process-step.ts
sriramasdanturthi: Now comes from the stepper through Lit context; I also removed the unused `last` property.

packages/forge/src/lib/core/styles/tokens/process-stepper/process-step/_tokens.scss
sriramasdanturthi: Removed; the ring uses the focus indicator's own tokens, which consumers can override through `::part(focus-indicator)`.

packages/forge/src/lib/core/styles/tokens/process-stepper/process-step/_tokens.scss
sriramasdanturthi: Now `border.variable(medium)`, which takes the line from 3px to 2px.

packages/forge/src/lib/core/styles/tokens/process-stepper/process-step/_tokens.scss
sriramasdanturthi: Done: `module-ref` to `line-width`.

packages/forge/src/lib/core/styles/tokens/process-stepper/process-step/_tokens.scss
sriramasdanturthi: Done.

samrichardsontylertech:
Looks good! I'm just going to wait to hear back from the design team before approving

PR #1252, opened 2026-09-18
chore: scaffold @tylertech/forge-rich-text-editor

derekmoss:
1 of 19 PRs
final destination is `feat/package-rte`

---

High level PR groups (more details will be provided in the follow up PRs for each concern)

pr01:  package scaffold

pr01b:   lift the rte from forge-extended verbatim

pr02 - pr13:  Port - use best of both worlds (forge and forge-extended) for new packages.

pr14 - pr19:  Fix issues found after enabling storybook.

---

This PR sets up the builds for rte only and does not contain any src code. (the build will fail only for this PR)


commitlint.config.mjs
derekmoss, line 1: The rename supposedly fixes conventional commit enforcement. (NOT TESTED)

according to clanker, the package.json does not have a "type" of "module".

 this means that the config is loaded as CommonJS instead of ESM and resolves to an empty ruleset.

arguably this would effect everything in this repo so not sure if we want to fix this here or not, (if it is an issue).

samrichardsontylertech (approved):

PR #1258, opened 2026-09-18
feat: add @tylertech/forge-react-rich-text-editor

derekmoss:
```

                                                                      ^ end of the port
pr19 ARIA  ← the fixes

reviewing changes to.
```

React wrappers around the editor's elements, generated by pr05's codegen.

**Review notes**

- **Generated output is not committed here**, unlike `@tylertech/forge-react`, which commits
  `src/components/component-wrappers.tsx`. `build` regenerates into the gitignored `dist/`, so
  `scripts/generate-proxies.mjs` is config only and the only file to edit when the surface changes.
  Worth a maintainer opinion on whether that asymmetry is acceptable long term.
- **Peer ranges are copied from `@tylertech/forge-react`** (`react >=17.0.0`) rather than
  independently derived, and are tested against one version, not the whole range.
- `tsconfig.build.json` type-checks generated declarations with **`skipLibCheck` off** deliberately.
  With it on, TypeScript skips declaration files entirely and the check passes even when a generated
  import resolves to nothing — which is the state `@tylertech/forge-extended-react` ships in today,
  where every generated declaration imports a nonexistent export.
- **Installing a pre-release tarball needs `--legacy-peer-deps`**: the adapter peers
  `@tylertech/forge-rich-text-editor@^0.1.0`, but `pnpm pack` produces `0.0.0` until the release
  bump. Only affects local tarball testing, not consumers.

---


samrichardsontylertech:
Should this be `forge-react-rich-text-editor` or `forge-rich-text-editor-react`? I'm going to throw this to the team to discuss

samrichardsontylertech (changes requested):
Looks good, the only change I'll ask for is to rename the package to `forge-rich-text-editor-react`

derekmoss:
Renamed to `@tylertech/forge-rich-text-editor-react`, with the directory moved to `packages/forge-rich-text-editor-react` to match. Folded into the original commit rather than added on top, so the package reads as correctly named from the start.

The rename also reached a few things beyond `package.json`: the changeset, the `pnpm-lock.yaml` importer key, the package's README and CLAUDE.md, and the dev harness added further up the stack. Verified nothing references the old name — no paths, no file contents, no commit messages, apart from the commit body explaining the rename itself.

Re-verified after the change: 19 React wrappers regenerate, 22 tests pass, builds and `format:check` clean, lint 0 errors.

Also rebased onto the merged base now that #1252–#1257 have landed.

samrichardsontylertech (approved):

PR #1217, opened 2026-08-31
Feat/forge extended migration

nickonometry:
## Summary
Migrates components (implementation, tests, Storybook stories/docs) from `@tylertech/forge-extended` into `@tylertech/forge`, converting each to this repo's Lit conventions (`BaseLitElement`, `setDefaultAria`, design tokens, side-effect dependency imports) along the way.

Any component still shipping from `@tylertech/forge-extended` is wired via an explicit subpath-only import (`@tylertech/forge/<name>`) rather than the root barrel, to avoid a duplicate `customElements.define()` collision for consumers with both packages loaded. These are tracked in `SUBPATH_ONLY_ENTRIES` in `packages/forge/scripts/build-esm.js`, which ensures each still gets its own `esm/<name>/index.js` output despite being excluded from the main barrel's Rollup entry graph.

Each migrated component also gets a blocks gallery demo (or several, covering additional story variants) under `blocks/src/blocks/components/<name>/`, and `blocks/forge-register.ts` is updated to source it from `@tylertech/forge` instead of `@tylertech/forge-extended` where applicable.

### Components migrated
- [x] `forge-busy-indicator`
- [x] `forge-footer` / `forge-footer-item`
- [x] `forge-app-layout`
- [x] `forge-app-launcher` / `forge-app-launcher-link`
- [x] `forge-confirmation-dialog`
- [x] `forge-count-card`
- [x] `forge-multi-select-header`
- [x] `forge-quantity-field`
- [x] `forge-responsive-toolbar`
- [x] `forge-structured-card` (+ its hard dependency `forge-content-scaffold`)
- [x] `forge-user-profile` / `forge-profile-link` (+ its hard dependency `forge-theme-toggle`)

All planned components are migrated.

## Checklist
- [x] Tests added/updated
- [x] Docs updated (if applicable)
- [x] Changeset added (`pnpm changeset`)

## Breaking Changes
None

PR #1259, opened 2026-09-18
feat: add @tylertech/forge-angular-rich-text-editor

derekmoss:
```

                                                                      ^ end of the port
pr19 ARIA  ← the fixes

reviewing changes to.
```

Angular component proxies and `NgModule`s around the editor's elements, generated by pr05's codegen
via `@tylertech/forge-schematics`.

**Review notes**

- **Generated proxies are committed here**, matching `@tylertech/forge-angular`. Do not hand-edit;
  regenerate.
- **The generator runs once per entry point.** The schematic accepts a single `importPath`, and the
  editor package exports core elements from the root and features from `./features`, so it runs two
  passes and merges. The schematic also emits no directory barrels and never touches the public API
  barrel (a TODO in its own source), so the codegen writes both plus an aggregate `ForgeRteModule`.
  Leaving that to hand maintenance is how the RTE proxies in `@tylertech/forge-extended-angular`
  ended up generated but unreachable. **If `forge-schematics` gains import-path overrides, the
  two-pass workaround can go away** — flagging because #1217 touches that package.
- **This PR declares `sass` and `vite`, and that is load-bearing.** `sass` is an optional peer of
  vite, and vite a peer of vitest. A package declaring `vite` or `@angular/build` without a matching
  `sass` forks a second vitest instance, and pnpm dedupes one shared `@angular/build` whose vitest
  link then points at the wrong one — so the `unit-test` builder runs on a different vitest runtime
  than the test project and the suite deadlocks after reporting exactly one spec file. This surfaced
  as **`@tylertech/forge-angular`'s existing suite hanging**, not this package's, which is why a new
  package appears to fix an unrelated one. Both must stay on the same catalog entries as every other
  vite-consuming package. The package CLAUDE.md records the diagnostic.
- **Peer ranges copied from `@tylertech/forge-angular`** (`>=20.0.0 < 23.0.0`), tested on one
  version.
- Pre-release tarball installs need `--legacy-peer-deps`, as in pr06.

---


samrichardsontylertech (changes requested):
This one looks good, just going to ask for the name to change to `forge-rich-text-editor-angular`

derekmoss:
Renamed to `@tylertech/forge-rich-text-editor-angular`, with the directory moved to `packages/forge-rich-text-editor-angular` to match. Folded into the original commit so the package reads as correctly named from the start.

Beyond `package.json` this also updated the Angular CLI project names in `angular.json` (library and demo), the `generate-proxies` config, `public-api.ts`, the tsconfigs, the `pnpm-lock.yaml` importer key, the package README and CLAUDE.md, and `forge-adapter-codegen`'s README, whose usage example names this package.

Re-verified after the change: proxies regenerate, 6 tests pass, builds and `format:check` clean, lint 0 errors.

Also rebased onto the merged base now that #1252–#1257 have landed.

samrichardsontylertech (approved):

PR #1260, opened 2026-09-18
feat: add dev harnesses for the editor and both adapters

derekmoss:
```

                                                                      ^ end of the port
pr19 ARIA  ← the fixes

reviewing changes to.
```

Scratch pages for looking at the components in a real browser: the editor on **3460**, React on
**3461**, Angular on **3462**. Plus Angular proxy smoke tests.

**Review notes**

- The harnesses import **built `esm/` rather than source**, so they exercise the same artifact a
  consumer gets. None of them are published.
- **Carries the React package's `sass` declaration**, for the reason spelled out in pr07. It is
  declared even though that package has no stylesheets of its own, so it looks like an unused
  dependency and is not one.
- `cli.analytics: false` in `angular.json` matches `@tylertech/forge-angular`'s existing convention
  and stops the CLI prompting in CI.
- Also drops a stray build dependency on forge's sass output and keeps compiled specs out of the
  published Angular package.

---


packages/forge-angular-rich-text-editor/README.md
```diff
@@ -53,3 +53,16 @@ It then writes the per-directory `index.ts` barrels, `src/public-api.ts` and the
...
+### Dev harness
+
+```bash
+pnpm dev   # build, then serve a scratch page at localhost:3462
+```
+
+`src/dev` is a page for looking at the components in a real browser. It is never published, and it
+imports built output rather than source, so it exercises the same artifact a consumer gets.
+
+Note that `forge-rich-text-editor` takes `content` as an **HTML string**, while
```
derekmoss, line 66: Validate, it should also be Prose Mirror Json

packages/forge-react-rich-text-editor/README.md
```diff
@@ -44,3 +44,16 @@ export, or if an attribute cannot be matched to a public property. The type chec
...
+### Dev harness
+
+```bash
+pnpm dev   # build, then serve a scratch page at localhost:3461
+```
+
+`src/dev` is a page for looking at the components in a real browser. It is never published, and it
+imports built output rather than source, so it exercises the same artifact a consumer gets.
+
+Note that `forge-rich-text-editor` takes `content` as an **HTML string**, while
```
derekmoss, line 57: Same as other

packages/forge-rich-text-editor/README.md
```diff
@@ -23,6 +23,22 @@ import '@tylertech/forge-rich-text-editor';
...
+pnpm dev     # serve a scratch page at localhost:3460
+pnpm test    # 529 tests in a real browser
+pnpm build   # rollup ESM, custom elements manifest and types
+```
+
+`src/dev` is a page for looking at the components in a real browser. It is never published, and it
+imports the built `esm/` output rather than source, so it exercises the same artifact a consumer
+gets.
+
+Note that `forge-rich-text-editor` takes `content` as an **HTML string**, while
```
derekmoss, line 38: same as other

packages/forge-angular-rich-text-editor/README.md
derekmoss: this is handled in the next PR in the stack.

packages/forge-react-rich-text-editor/README.md
derekmoss: this is handled in the next PR in the stack.

packages/forge-rich-text-editor/README.md
derekmoss: this is handled in the next PR in the stack.

samrichardsontylertech (approved):

PR #1268, opened 2026-09-18
fix: make the disabled and readonly states visible

derekmoss:
```

                                                                      ^ end of the port
pr19 ARIA  ← the fixes

reviewing changes to.
```

Neither state dimmed anything, so a disabled editor was indistinguishable from an editable one.

**Review notes**

- Root cause was a reference to `--forge-theme-disabled-opacity`, **which is not a forge token**, so
  the rule resolved to nothing. Now uses `forge-theme.emphasis(medium-low)` behind a
  `--forge-rich-text-editor-disabled-opacity` custom property.
- The content element's disabled/readonly rules were dead code and are removed.
- Disabled text was still selectable: `pointer-events: none` does not imply unselectable.
- An existing spec carried the comment *"Verify the component has disabled attribute which triggers
  `:host([disabled])` styles"* with **no style assertion at all** — it asserted the attribute it had
  just set. The new tests assert computed style.
- Flagging for #1184: this and pr14 call `forge-theme.emphasis()`, which resolves through
  `color-emphasis.value()` and `math.div`.

---


packages/forge-rich-text-editor/src/lib/rich-text-content.scss
```diff
@@ -49,8 +50,9 @@
 
 :host([disabled]) {
   .editor-content-wrapper {
-    opacity: var(--forge-theme-disabled-opacity);
+    opacity: var(--forge-rich-text-content-disabled-opacity, #{forge-theme.emphasis(medium-low)});
```
derekmoss, line 53: this probably should not have been changed from `--forge-theme-disabled-opacity`

packages/forge-rich-text-editor/src/lib/rich-text-editor.scss
```diff
@@ -49,14 +50,18 @@
 
 :host([disabled]) {
   .forge-rich-text-editor {
-    opacity: var(--forge-theme-disabled-opacity);
+    opacity: var(--forge-rich-text-editor-disabled-opacity, #{forge-theme.emphasis(medium-low)});
```
derekmoss, line 53: same as previous comment it should just be `--forge-theme-disabled-opacity`

packages/forge-rich-text-editor/src/lib/rich-text-editor.ts
```diff
@@ -46,6 +46,9 @@ export const RICH_TEXT_EDITOR_TAG_NAME: keyof HTMLElementTagNameMap = 'forge-ric
  * @property {boolean} [allowPasteImages=false] - Whether to allow images to be pasted into the editor.
  *
  * @attribute {string} content - The HTML content of the editor. Pass a ProseMirror document through the property instead; an attribute cannot carry an object.
+ * @cssproperty --forge-rich-text-editor-disabled-opacity - The opacity of the editor when it is
```
derekmoss, line 49: not sure this is needed

packages/forge-rich-text-editor/src/lib/rich-text-content.scss
derekmoss: seems like this is already the pattern in forge to have per component disabled-opacity with falling back to a more general shared opacity. dismissing my concern.

packages/forge-rich-text-editor/src/lib/rich-text-editor.scss
derekmoss: seems like this is already the pattern in forge to have per component disabled-opacity with falling back to a more general shared opacity. dismissing my concern.

packages/forge-rich-text-editor/src/lib/rich-text-editor.ts
derekmoss: seems like this is already the pattern in forge to have per component disabled-opacity with falling back to a more general shared opacity. dismissing my concern.

samrichardsontylertech (approved):

PR #1271, opened 2026-09-18
fix: correct three ARIA defects, with an axe suite

derekmoss:
```

                                                                      ^ end of the port
pr19 ARIA  ← the fixes

reviewing changes to.
```

Adds an axe-core suite and fixes what it found.

1. **The editable element had no accessible name.** Tiptap builds its own `contenteditable` inside
   the element it is handed, and that inner element is the real `textbox` — focusable, and where
   assistive technology lands. The name sat on the container, so the field a user actually reaches
   was anonymous. Now named through `editorProps`.
2. **`aria-controls` referenced an id across a shadow boundary.** The toolbar and all thirteen tool
   buttons carried `aria-controls="forge-rte-content"`, but that id lives inside
   `forge-rich-text-content`'s shadow root, and an IDREF cannot cross one. All fourteen references
   were invalid, which axe rates critical. **The attribute is removed rather than repointed** —
   there is no same-root element to reference, and cross-root ARIA is not broadly supported. Worth
   a maintainer opinion on that call.
3. **The renderer advertised itself as a text input.** ProseMirror marks its element
   `role="textbox"` even when not editable, leaving the renderer an unnamed read-only text field.
   The host already takes `role="article"`, so the inner element now carries `role="presentation"`
   while its children keep their own semantics.

The suite covers the editable, readonly, disabled, counted and error states, the link popover while
open, a composed layout, and the renderer. **Reverting the three fixes fails all ten cases.**

**Review notes**

- **Two existing specs asserted the invalid `aria-controls` value as correct**, so they are changed
  here. Worth confirming you agree with the new expectation rather than assuming the old one was
  load-bearing.
- The popover case asserts the popover actually opened before running axe. The renderer case
  initially read clean **only because Tiptap had not mounted inside the settle window** — a check
  that would have passed forever without testing anything. Worth knowing if you add cases.
- `src/testing/a11y.ts` mirrors forge's own `core/testing/a11y.ts`, which is not reachable from here
  — forge exports no testing subpath. `expectNoA11yViolations` fails with the full violation list
  rather than a count, so a regression is readable without a rerun.
- **`axe-core` needed no new resolution**: it was already in the lockfile via `packages/forge` and
  `@storybook/addon-a11y`, so only an importer entry was added.
- **Unverified, and consistent with this repo's chromium-only posture:** Safari, and the behavior of
  axe in a CI automation environment as opposed to locally. Flagging rather than claiming coverage.
- Suite at this tip: **612 passing / 26 skipped**.

---


packages/forge-rich-text-editor/src/lib/features/core/rte-tool-button.ts
```diff
@@ -89,8 +89,7 @@ export class RteToolButtonComponent extends LitElement {
         @keydown=${this.#handleKeydown}
         ?disabled=${this.disabled}
         aria-label=${this.label}
-        aria-keyshortcuts=${this.keyboardShortcut || ''}
-        aria-controls="forge-rte-content">
```
samrichardsontylertech: This can be replaced with `ariaControlsElements` to cross the shadow boundary (https://developer.mozilla.org/en-US/docs/Web/API/Element/ariaControlsElements)

packages/forge-rich-text-editor/src/lib/rich-text-renderer.ts
```diff
@@ -136,6 +136,16 @@ export class RichTextRendererComponent extends LitElement {
...
           extensions: DEFAULT_EXTENSIONS,
           editable: false,
+          // ProseMirror marks its element `role="textbox"` even when it is not editable, which left
+          // the renderer advertising an unnamed text input - axe reports it as
+          // aria-input-field-name. This is display surface, not a form field, and the host already
+          // carries `role="article"`, so the inner element should carry no role of its own. Its
+          // children keep their own semantics.
+          editorProps: {
+            attributes: {
+              role: 'presentation'
```
samrichardsontylertech: I wonder whether `aria-readonly: 'true'` would be appropriate here instead of removing the textbox role?

samrichardsontylertech (changes requested):

derekmoss:
Both applied, with one qualification each that I verified in the browser rather than assuming.

## `ariaControlsElements`

Good call — it's supported (Chrome 147 here) and it does cross a shadow boundary. But it only resolves into **the same tree or an ancestor tree**, and I measured the rest:

| target | result |
| --- | --- |
| the editable element (in `forge-rich-text-content`'s shadow root) | silently dropped |
| `forge-rich-text-content` host (wrapped layout) | silently dropped |
| the editor / context element | resolves |

The editable element sits in a sibling branch to the toolbar's, so pointing at it would have failed the same way the IDREF did — just without axe telling us. `ShadowRoot.referenceTarget`, which would let the content host stand in for its inner element, isn't available yet (`'referenceTarget' in ShadowRoot.prototype === false`), so there's no way to reach the editable element today.

So the reference targets **the editor**. The target is published on the editor context as `controlsElement`, resolved by the provider walking to its outermost host so it's right in both the wrapped and composed layouts. Feature elements pass it down, which keeps `forge-rte-tool-button` free of any context dependency — worth preserving, since a detached toolbar depends on that.

One thing to expect in the inspector: **Chrome reflects the property by writing an empty `aria-controls=""`.** It references nothing and axe reports no violation, so the suite now asserts the absence of a *non-empty* IDREF rather than the absence of the attribute. Four of my own tests failed on that first, which is how I found it.

If you'd rather not grow `EditorContext`, the alternative is for the button to resolve the editor itself via `closest()` after walking out of its shadow roots — one file instead of twelve, but it couples the button to the tag names. Happy to switch.

## `aria-readonly` on the renderer

Also a better description than removing the role — but it needs a name. I checked all three variants against axe:

| variant | axe |
| --- | --- |
| `role="presentation"` (what this PR did) | clean |
| `role="textbox" aria-readonly="true"` | **`aria-input-field-name`** |
| same + `aria-label` | clean |

Without a name it reintroduces the exact violation the third fix removed. So it's now `role="textbox"` + `aria-readonly="true"` + `aria-label="Rich text content"`, mirroring how the editor names its own textbox. If you'd prefer that name to be consumer-settable rather than fixed, that's a small property addition — say the word.

Suite is **614 passing / 26 skipped**, lint and format clean. The changeset now describes what actually landed rather than "the attribute is removed".

derekmoss:
not ready for review, I still had a Opus 5.0 thread going and it was very aggressive on git pushing and leaving comments.

packages/forge-rich-text-editor/src/lib/rich-text-renderer.ts
derekmoss: I am going to paraphrase with my limited Opus 5.5 accessibility research. 
Agent pushed back so I asked it why. For which at least in my head makes sense.

The rich-text-renderer is not a textbox, so the role should be like presentation.
Then the hierarchy becomes Article > Presentation > [ Heading#'s, lists, links]

For the rich-text-editor it makes sense to do aria-readonly="true", but not for the renderer.

